PR #157, opened 2026-04-14
fix: include closed tasks in progress bar completion count

drstrangerujn:
Closes #120

Progress bars at project and group level only counted `done` tasks as completed, but `closed` tasks still counted toward the total. So if you close a task instead of marking it done, progress goes *down* — which is confusing.

## What changed

- **`src/app/api/projects/route.ts`** — Prisma query now filters `status: { in: ["done", "closed"] }` instead of just `"done"`
- **`overview-timeline.tsx`** — task stats count `done || closed`
- **`task-list-view.tsx`** — same fix for the task list aggregate

This matches what `project.service.ts` already does (line 255):
```ts
const tasksDone = (taskStatusMap.get("done") || 0) + (taskStatusMap.get("closed") || 0);
```

## Test

1. Create a project with tasks
2. Close some tasks (not mark as done)
3. Progress bar should reflect closed tasks as completed

### DIFF
--- a/src/app/(dashboard)/projects/[uuid]/dashboard/panels/overview-timeline.tsx
+++ b/src/app/(dashboard)/projects/[uuid]/dashboard/panels/overview-timeline.tsx
@@ -185,7 +185,7 @@ export function OverviewTimeline({
 
   // Task stats
   const taskStats = useMemo(() => {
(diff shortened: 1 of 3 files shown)


ChenNima (approved):
LGTM, merging it

PR #158, opened 2026-04-14
fix: allow reassigning already-assigned tasks

### LINKED ISSUE #12: Reassign task fails and button is pushed off-screen
## Bug

当已 assign agent 的任务想 reassign 时:
1. Reassign 失败
2. 按钮被推到面板底部，无法点击

## Expected
- Reassign 应该对 open/assigned 状态的任务生效
- 按钮应始终可点击

## Priority
P1 - 影响正常任务管理流程

Reported by Junjie in Slack (2026-03-05)

drstrangerujn:
Closes #12 (reassign part — button layout is a separate CSS issue)

## What's wrong

`claimTask()` in `task.service.ts` only matches tasks with `status: "open"`:

```ts
where: { uuid: taskUuid, status: "open" }
```

But the UI-side check in `actions.ts` allows both `open` and `assigned`:

```ts
if (task.status !== "open" && task.status !== "assigned") {
  return { success: false, error: "Task is not available for claiming" };
}
```

So when you try to reassign a task that's already assigned to someone, the UI says "go ahead" but the service layer says "nope" — Prisma can't find a matching row and throws P2025, which becomes `AlreadyClaimedError`.

## Fix

```ts
where: { uuid: taskUuid, status: { in: ["open", "assigned"] } }
```

Now reassigning works. If the task is in any other status (in_progress, to_verify, done, closed), it still correctly rejects.

## Tests

- Updated existing test assertion to match the new where filter
- Added a new test case: "reassigns an already-assigned task to a different agent"
- All 81 tests pass

### DIFF
--- a/src/services/__tests__/task.service.test.ts
+++ b/src/services/__tests__/task.service.test.ts
@@ -482,7 +482,7 @@ describe("claimTask", () => {
     expect(result.status).toBe("assigned");
     expect(mockPrisma.task.update).toHaveBeenCalledWith(
       expect.objectContaining({
-        where: { uuid: TASK_UUID, status: "open" },
+        where: { uuid: TASK_UUID, status: { in: ["open", "assigned"] } },
         data: expect.objectContaining({
           status: "assigned",
           assigneeType: "agent",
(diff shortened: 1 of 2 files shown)


ChenNima (approved):
LGTM. Real bug — UI and MCP callers allow `assigned` status but service layer only matched `open`. Fix is correct and minimal. Tests look good.

Minor nits (non-blocking):
- `task.service.ts:589` comment still says "only succeeds if status is open" — could update to include "assigned"
- `developer.ts:20` MCP tool description "(open -> assigned)" could mention reassign

PR #232, opened 2026-05-02
feat(authz): fine-grained agent permissions (0.7.0)

ChenNima:
## Summary

Replaces the fixed 3-role agent model (admin/pm/developer) with a fine-grained permission matrix: **5 resources × 3 actions = 15 bits**. The UI surfaces this as 3 role presets plus a **Custom** option; both onboarding and settings now share a single form component, and Create/Edit use the same inner fields.

The goal is flexibility: agents can be scoped as narrowly or broadly as the deploying user needs (e.g. a PM agent that can also verify its own tasks, or a read-only auditor agent).

## Highlights

- **authz library** (`src/lib/authz/`): types, presets, and `computeEffectivePermissions(roles, custom)` with full test coverage
- **REST + MCP gating**: `requireAgentPermission` for routes, `registerPermissionedTool` / `permission-map.ts` for MCP tools
- **Checkin schema**: returns a resource-aggregated object (`{ idea: ["read","write","admin"], ... }`) — token-efficient for skills and plugin hooks
- **UI**: `AgentFormFields` shared by onboarding, Settings create modal, Settings edit modal. `AgentPermissionPicker` shows a 5×3 grid with a Custom option; the Admin column is always interactive (toggling it on a preset simply flips the form to Custom with the new bit included)
- **Plugin 0.8.0** + standalone skills 0.2.1/0.1.1: all prereq checks rewritten against the new permission shape; session-start hook extracts `agent_permissions` instead of `agent_roles`
- **Migration**: adds `permissions` column on Agent (ApiKey inherits permissions via its Agent); legacy roles continue to work as preset identifiers

## Changed files (groups)

| Area | What changed |
|------|--------------|
| `src/lib/authz/*` | New permission types, presets, effective-set computation |
| `src/services/{checkin,agent}.service.ts` | Aggregated permission output; permission-aware agent CRUD |
| `src/mcp/**` | Per-tool permission declarations; new `permission-map.ts` (test coverage contract) and `register-helpers.ts` |
| `src/app/api/**` | All REST routes gated via `requireAgentPermission`; claim routes updated for the new `*_agent` stack |
| `src/app/(dashboard)/settings/page.tsx` | Edit modal uses shared `AgentFormFields`; preset derivation considers custom permissions so extras aren't silently dropped; admin-warning section removed |
| `src/components/AgentFormFields.tsx` (new) | Shared inner form (name + permission picker + persona) |
| `src/components/AgentCreateForm.tsx` | Uses `AgentFormFields`; supports embedded mode for onboarding |
| `src/components/AgentPermissionPicker.tsx` | 5×3 grid + preset selector incl. Custom; admin column always interactive |
| `src/app/onboarding/components/CreateAgentStep.tsx` | Wraps `AgentCreateForm embedded` — no duplicate form logic |
| `public/chorus-plugin/*` | Bumped to 0.8.0; hooks + skills switched to permission-based checks |
| `public/skill/*` | Standalone skill docs updated to aggregated permission model |
| `prisma/migrations/...add_agent_permissions` | Adds `permissions` column on Agent |
| `docs/{MCP_TOOLS,design.pen}` | Documentation + design.pen reflect the new permission UI |

## Review fixes (second commit)

- Settings edit modal preset derivation now respects stored custom permissions (blocker: extras were silently dropped on save).
- Claim routes (`/api/ideas|tasks/[uuid]/claim`) stopped filtering by legacy `pm` / `developer` role strings — they gate on `idea:write` / `task:write` of the selected agent instead, so any 0.7.0 agent is eligible.
- `chorus_pm_{reject,revoke}_proposal` use `hasPermission(auth, "proposal:admin")` instead of role-string checks.
- `POST /api/agents` drops legacy pm/developer/admin aliases to match PATCH.
- `/api/mcp` widens the AgentAuthContext roles cast to the full AgentRole union.

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] `pnpm test` — 1464 pass, 1 skipped
- [x] `pnpm test src/components` — AgentPermissionPicker contract tests updated for always-interactive Admin column
- [x] `pnpm test src/lib/authz` — permission computation fully covered
- [x] New integration test: `src/__tests__/integration/permissions.test.ts`
- [x] New API gating test: `src/app/api/__tests__/permission-gating.test.ts`
- [x] New MCP server test: `src/mcp/__tests__/server.test.ts`
- [x] Manual UI: onboarding → Settings create → Settings edit all render identical fields via the shared component; admin column toggles freely on any preset; no admin-warning popup in edit modal

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -11,7 +11,7 @@
       "name": "chorus",
       "source": "./public/chorus-plugin",
       "description": "Chorus AI-DLC collaboration platform plugin. Automates session lifecycle, provides MCP tools for PM/Developer/Admin workflows, and enables multi-agent team observability.",
-      "version": "0.7.5",
+      "version": "0.8.0",
       "category": "project-management",
       "tags": ["ai-dlc", "collaboration", "mcp", "session", "multi-agent"]
     }
--- a/docs/MCP_TOOLS.md
+++ b/docs/MCP_TOOLS.md
@@ -4,13 +4,81 @@ This document covers all tools provided by the Chorus MCP Server, including tool
 
 ## Overview
 
-The Chorus MCP Server provides different tool sets based on Agent roles:
+Tool visibility is driven by a **fine-grained permission model**: 5 resources (`idea`, `proposal`, `document`, `task`, `project`) × 3 actions (`read`, `write`, `admin`) = **15 permissions**. Each gated tool declares a single **required permission**. Public tools (discover, comment, session) carry no gate and are always available.
+
+Agents may use a **role preset** (`developer_agent`, `pm_agent`, `admin_agent`) that expands to a fixed permission set, and/or **custom permissions** added on top. The effective permission set is the union of preset + custom. See `src/lib/authz/presets.ts` for the authoritative preset mapping, `src/mcp/tools/permission-map.ts` for the tool → permission map, and [ARCHITECTURE.md §6.3](./ARCHITECTURE.md#63-permission-model) for the conceptual overview.
+
+### Role Preset → Permission Set
+
+| Preset | Permission Set |
+|--------|----------------|
+| `developer_agent` | `*:read` + `task:write` (6 perms) |
+| `pm_agent` | `*:read` + `idea:write`, `proposal:write`, `document:write`, `task:write`, `project:write` (10 perms) |
+| `admin_agent` | all 15 perms (`*:read` + `*:write` + `*:admin`) |
+
+Read-only tools (`chorus_get_*`, `chorus_list_*`, `chorus_checkin`, `chorus_search*`, comments, elaboration answers, session management, `chorus_create_tasks`, `chorus_update_task`) are **public** and available to any agent regardless of preset/permissions — they are listed under "Public Tools" and "Session Tools" below without a Required Permission row.
+
+> Note: `chorus_create_tasks` and `chorus_update_task` field edits are public because handler-level assignee / authorship guards enforce who can actually mutate state. Operational status transitions (`in_progress`, `to_verify`) still require the caller to be the task's assignee at the service layer — the permission gate is about tool visibility, not operation authorization.
+
+### Gated Tool → Required Permission Matrix
+
+The following table summarizes every permission-gated MCP tool. Each tool has exactly one required permission; possessing that permission (via preset or custom) is the **necessary** condition for the tool to appear in the agent's tool list. Additional handler-level guards (ownership, assignee, status) may still apply.
+
+| Tool | Required Permission |
+|------|---------------------|
+| `chorus_claim_idea` | `idea:write` |
+| `chorus_release_idea` | `idea:write` |
+| `chorus_move_idea` | `idea:write` |
+| `chorus_pm_create_idea` | `idea:write` |
+| `chorus_pm_start_elaboration` | `idea:write` |
+| `chorus_pm_validate_elaboration` | `idea:write` |
+| `chorus_pm_skip_elaboration` | `idea:write` |
+| `chorus_pm_create_proposal` | `proposal:write` |
+| `chorus_pm_validate_proposal` | `proposal:write` |
+| `chorus_pm_submit_proposal` | `proposal:write` |
+| `chorus_pm_add_document_draft` | `proposal:write` |
+| `chorus_pm_add_task_draft` | `proposal:write` |
+| `chorus_pm_update_document_draft` | `proposal:write` |
+| `chorus_pm_update_task_draft` | `proposal:write` |
+| `chorus_pm_remove_document_draft` | `proposal:write` |
+| `chorus_pm_remove_task_draft` | `proposal:write` |
+| `chorus_pm_reject_proposal` | `proposal:write` |
+| `chorus_pm_revoke_proposal` | `proposal:write` |
+| `chorus_pm_create_tasks` | `proposal:write` |
+| `chorus_add_task_dependency` | `proposal:write` |
+| `chorus_remove_task_dependency` | `proposal:write` |
+| `chorus_pm_assign_task` | `proposal:write` |
+| `chorus_pm_create_document` | `document:write` |
+| `chorus_pm_update_document` | `document:write` |
+| `chorus_claim_task` | `task:write` |
+| `chorus_release_task` | `task:write` |
+| `chorus_submit_for_verify` | `task:write` |
+| `chorus_report_criteria_self_check` | `task:write` |
+| `chorus_report_work` | `task:write` |
+| `chorus_admin_create_project` | `project:write` |
+| `chorus_admin_create_project_group` | `project:write` |
+| `chorus_admin_update_project_group` | `project:write` |
+| `chorus_admin_delete_project_group` | `project:write` |
+| `chorus_admin_move_project_to_group` | `project:write` |
+| `chorus_admin_approve_proposal` | `proposal:admin` |
+| `chorus_admin_close_proposal` | `proposal:admin` |
+| `chorus_admin_verify_task` | `task:admin` |
+| `chorus_admin_reopen_task` | `task:admin` |
+| `chorus_admin_close_task` | `task:admin` |
+| `chorus_mark_acceptance_criteria` | `task:admin` |
+| `chorus_admin_delete_task` | `task:admin` |
+| `chorus_admin_delete_idea` | `idea:admin` |
+| `chorus_admin_delete_document` | `document:admin` |
+
+### Legacy Role → Tool Set View
+
+For agents that rely on the preset alone (no custom permissions), this is the resulting tool set:
 
 | Role | Tool Set |
 |------|----------|
-| Developer Agent | Public + Session + Developer |
-| PM Agent | Public + Session + PM |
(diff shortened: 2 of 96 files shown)


ChenNima (commented):
## Code review — `feat(authz): fine-grained agent permissions (0.7.0)`

Overall: the core authz library (`types`/`presets`/`permissions`), the MCP `registerPermissionedTool` gate, and the test coverage (`server.test.ts`, `integration/permissions.test.ts`, `AgentPermissionPicker.test.tsx`) are in great shape. The migration is a clean additive `ALTER TABLE`, presets preserve 0.6.x behavior, and the resource-aggregated checkin shape is a nice token-efficient win. The items below are what stood out that I would want addressed (or at least explicitly acknowledged) before merging.

---

### 🔴 Likely data-loss in the Settings edit modal

`src/app/(dashboard)/settings/page.tsx:164-172` + `src/components/AgentFormFields.tsx:50-53`:

- `derivePresetFromRoles(key.roles)` returns the preset key whenever `roles.length === 1 && roles[0] in ROLE_PRESETS`.
- When `preset !== \"custom\"`, `AgentFormFields.pickerPermissions` is `[...ROLE_PRESETS[preset]]` — the stored `permissions` array is **not surfaced** in the picker.
- On save (page.tsx:197-199), preset mode persists `permissions: []`.

So an agent with `roles: [\"pm_agent\"]` **and** non-empty `permissions` (e.g. someone manually granted `task:admin` on top of the preset) will (a) not show the extra bit in the edit UI, and (b) lose it on any save — silently.

The new create/onboarding flows always emit *either* `roles` *or* `permissions`, never both, so the risk window is agents that either already exist in DB with mixed state or are created via a future API-first flow. Suggested fix in `derivePresetFromRoles`:

```ts
function derivePreset(roles: string[], permissions: string[]): EditPreset {
  if (permissions.length > 0) return \"custom\";           // preserve custom bits
  if (roles.length === 1 && roles[0] in ROLE_PRESETS) return roles[0] as PresetKey;
  return \"custom\";
}
```

…and when `preset === \"custom\"` initialize `editPermissions` with `effectivePermissions` (already done) so a round-trip is lossless.

---

### 🟡 Role check leaks into an otherwise permission-based system

`src/mcp/tools/pm.ts:1039` + usage at 1060 / 1110:

```ts
const hasAdminRole = auth.roles.some(r => r === \"admin\" || r === \"admin_agent\");
...
if (!hasAdminRole && proposal.createdByUuid !== auth.actorUuid) { ... }
```

`chorus_pm_reject_proposal` and `chorus_pm_revoke_proposal` let admins act on any proposal, but \"admin\" is determined by the legacy role string, not by the new permission bits. A custom agent with `proposal:admin` granted explicitly is rightly visible this tool (gated on `proposal:write`, which `proposal:admin` is typically paired with in custom setups), but would fail the ownership check — arguably the *more* capable custom agent gets less capability than a preset admin. Consistency would suggest:

```ts
const canActOnAny = hasPermission(auth, \"proposal:admin\");
```

---

### 🟡 Legacy role filter in user claim routes will miss all newly-created agents

`src/app/api/ideas/[uuid]/claim/route.ts:54` and `src/app/api/tasks/[uuid]/claim/route.ts:54`:

```ts
roles: { has: \"pm\" },         // claim idea
roles: { has: \"developer\" },  // claim task
```

Everything else in the stack (`ROLE_PRESETS`, `AgentCreateForm` default `[\"admin_agent\"]`, `/api/agents` POST validator) uses the `*_agent` suffixed form. `computeEffectivePermissions` tolerates both (`normalized = role.endsWith(\"_agent\") ? role : \`${role}_agent\``) but the Prisma `has:` filter does not. Result: a user trying to assign an idea/task to one of their agents via the modal will see \"PM Agent not found\" / \"Developer Agent not found\" for any agent created through the new onboarding or settings flow.

This may be pre-existing — but the PR actively widens the window by standardizing agent creation on the `*_agent` form. The fix is a two-key filter (`OR: [{ roles: { has: \"pm\" } }, { roles: { has: \"pm_agent\" } }]`) or, better, a permission check (`permissions: { hasSome: [\"idea:write\"] }` / `[\"task:write\"]` — but note Prisma doesn't index array overlap against the preset-expanded set, so this may need an in-memory filter via `computeEffectivePermissions`).

---

### 🟡 POST vs PATCH role validator asymmetry

- `src/app/api/agents/route.ts:97` (POST) accepts `[\"pm_agent\", \"developer_agent\", \"admin_agent\", \"pm\", \"developer\", \"admin\"]`.
- `src/app/api/agents/[uuid]/route.ts:122` (PATCH) accepts only `[\"pm_agent\", \"developer_agent\", \"admin_agent\"]`.

A client that legitimately POSTs with a legacy role cannot PATCH it back. Pick one and enforce it in both places — or, since you're accepting legacy roles, also accept them on PATCH for symmetry.

---

### 🟡 `permission-map.ts` is test-only; runtime uses inline gates

The module comment in `src/mcp/tools/permission-map.ts` calls it the \"Central mapping of MCP tools to their required Permission\" but the three tool modules pass their required permission inline as the 3rd arg to `registerPermissionedTool`. The only production importer is `src/mcp/__tests__/server.test.ts`. The coverage test (`registered === Object.keys(TOOL_PERMISSIONS)`) catches drift, but you still have to keep two places in sync. Either:
- Make `permission-map.ts` the source of truth and have `registerPermissionedTool` look up by name (`tool_name → required`), or
- Drop the map and derive it from the registration closures at test-time only (collect `{ name, required }` pairs from the registrars).

Keeping the status quo is fine — just worth a one-line comment clarifying the map is a test-only manifest, or you'll eventually see a PR that edits one and forgets the other.

---

### 🔵 Minor

- **PR description vs migration**: \"adds `permissions` column on Agent/ApiKey\" — migration only touches Agent (`prisma/migrations/20260430034110_add_agent_permissions/migration.sql`). Functionally fine (ApiKey resolves to its Agent), but the description should match.
- **`src/app/api/mcp/route.ts:59`** casts `validation.agent.roles as (\"pm\" | \"developer\" | \"admin\")[]` — with the `*_agent` forms now standard, this cast is literally wrong at runtime. Should be the broader `AgentRole` union defined in `src/types/auth.ts`.
- **Super_admin behavior** is intentional per the integration test but the two gate helpers in `src/lib/auth.ts` (`checkAgentPermission` vs `requireAgentPermission`) handle super_admin differently. A one-line doc comment on each saying which is bypass-on-super-admin and which is not would save future confusion.
- **`src/lib/authz/permissions.ts:53`** — the uncovered branch (`continue` on invalid permission) is reachable only via direct DB manipulation; if you want that covered, a unit test that feeds `computeEffectivePermissions(null, [\"bogus:garbage\"])` would do it.

---

### ✅ What I liked

- `computeEffectivePermissions` gracefully handles both `\"pm\"` and `\"pm_agent\"` — good back-compat defense.
- `groupPermissionsByResource` canonical sort (read, write, admin) gives stable checkin output — nice for skills that substring-match.
- `registerPermissionedTool` is refreshingly small: one `if` + delegate. No magic.
- `AgentPermissionPicker` tests cover the subtle \"admin column interactive on preset flips to custom\" behavior — that's exactly the kind of UI contract worth pinning.
- `integration/permissions.test.ts` \"+10 diff\" comment documents a prior decision thread — future-you will thank you.

Nothing blocking that I can't see being resolved with a follow-up; the data-loss one is the only thing I'd ask for before merge, and the role-based `hasAdminRole` is the strongest candidate for a \"fix now so nobody builds on it\" change.

ChenNima:
Re-review on `ef8d217`: all six items addressed, CI green. 👍

- **Settings edit preset** — `deriveEditPreset(roles, customPermissions)` requires both a single preset role AND empty custom perms; extras fall to Custom mode, so the round-trip is now lossless. Nice comment explaining the intent.
- **`pm_reject`/`pm_revoke`** — now `hasPermission(auth, "proposal:admin")`. Consistent with `chorus_admin_approve_proposal`. The rename to `canAdminAnyProposal` also reads better than the old role-flavored name.
- **Claim routes** — nice call going with the in-memory permission check (vs. a two-key Prisma filter). Bonus: a developer-preset agent explicitly granted `idea:write` is now eligible for idea assignment, which the old role filter would have blocked. `errors.notFound("Agent")` + `errors.forbidden("Selected agent does not have X:write permission")` also gives a clearer error path for the UI.
- **POST `/api/agents`** — role allowlist now `[pm_agent, developer_agent, admin_agent]`, matches PATCH.
- **`/api/mcp` cast** — widened to `AgentRole[]`, correct now that `*_agent` is standard.
- **`permission-map.ts`** — header rewrite makes the test-only contract explicit. Exactly what I was hoping for.

Two small consequential observations (non-blocking, just noting for the record):

1. `permissions` + legacy string roles can no longer be created via POST, but they *can* still exist in DB (either from 0.6.x data or from direct seeding). `deriveEditPreset` correctly falls through to Custom for anything unusual, so the UI degrades gracefully. 👍
2. Claim-route behavior is a *small* widening vs. 0.6.x: a custom agent with `idea:write` but no `pm` role can now be picked as an idea assignee (and same for tasks/`task:write`). That's what "fine-grained permissions" is supposed to mean, so I think it's correct — but worth flagging in the release notes so users who edit their agents don't get surprised.

LGTM to ship.

PR #272, opened 2026-05-25
feat(sessions): active/closed lifecycle + lastActiveAt filter (0.9.0)

ChenNima:
## Summary
- Collapse `AgentSession.status` to `{active, closed}`; drop the unused `expiresAt` field; remove dead code (`markInactiveSessions`, `inactive→active` recovery branch, every `inactive` literal).
- Add a 1h `lastActiveAt` staleness filter applied **only** on default UI lists (Settings sessions list, project worker-avatar header). MCP reads (`chorus_list_sessions`, `chorus_get_session`), the REST `/api/sessions/[uuid]` route, and Activity-stream session-name lookups remain unfiltered so plugin reuse and audit-trail links keep working.
- Bake heartbeat into every session-touching MCP tool (`getSession`, `closeSession`, `reopenSession`, `sessionCheckinToTask`, `sessionCheckoutFromTask`, `heartbeatSession`) with a guard that skips the refresh when `status='closed'`. Plugins no longer need standalone heartbeats during normal operation.

Implements idea 8f39105e via OpenSpec change `session-lifecycle-active-only` (archived under `openspec/changes/archive/2026-05-25-…/`); new long-term capability spec at `openspec/specs/session-lifecycle/spec.md`.

## Changed files
| File | Change |
|------|--------|
| `prisma/schema.prisma` | Drop `expiresAt`; comment now says `active \| closed`; `@@index([lastActiveAt])` added |
| `prisma/migrations/20260524144057_remove_session_expires_at/migration.sql` | Pure DDL — `DROP COLUMN expiresAt` + `CREATE INDEX AgentSession_lastActiveAt_idx`. No DML per project policy; residual `inactive` rows are excluded by the filter. |
| `src/services/session.service.ts` | Add `SESSION_STALE_THRESHOLD_MS`, `freshSessionWhereClause`, `touchLastActiveAt` (with closed-guard); split `listAgentSessions` (unfiltered, MCP) vs new `listAgentSessionsForUI` (filtered); wire `touchLastActiveAt` into all session-touching paths; delete `markInactiveSessions`, `inactive` literals, and `inactive→active` heartbeat recovery branch |
| `src/mcp/tools/session.ts` | Drop `expiresAt` from `chorus_create_session`; drop `inactive` from `chorus_list_sessions` status enum |
| `src/app/(dashboard)/settings/actions.ts` | Server action now calls `listAgentSessionsForUI` |
| `src/app/api/agents/[uuid]/sessions/route.ts` | Route now calls `listAgentSessionsForUI`; doc comment clarifies UI-vs-MCP filtering split |
| `src/app/(dashboard)/settings/page.tsx` | Count badge drops the post-filter; remove `inactive` styling branches; closed-reopen branch left defensively |
| `messages/{en,zh}.json` | `sessions.noSessions` empty-state message updated to "No active sessions in the last hour" / "近 1 小时内无活跃会话" |
| `src/services/__tests__/session.service.test.ts` | +16 tests: 1h cutoff boundary, filter applied vs not, getActiveSessionsForProject + batchGetWorkerCountsForTasks shape, heartbeat side-effect for each tool, closed-session guard, closeSession does not mutate Task.status, regression guard against `inactive` reintroduction |
| `docs/MCP_TOOLS.md` | New "Lifecycle and staleness" + "Implicit-heartbeat contract" sections; per-tool descriptions updated |
| `CHANGELOG.md` | 0.9.0 Unreleased "Changed (Breaking)" entry |
| `openspec/specs/session-lifecycle/spec.md` | New long-term capability spec (8 ADDED Requirements with scenarios) |
| `openspec/changes/archive/2026-05-25-session-lifecycle-active-only/` | Archived OpenSpec change (proposal, design, delta spec, tasks, README) |

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `pnpm test src/services/__tests__/session.service.test.ts` — 48/48 pass
- [x] `pnpm test` full suite — 1588 pass / 1 skipped (pre-existing)
- [x] Coverage 95.91 lines / 88.55 branches (above 95/85 thresholds); session.service.ts at 99.07/98.14
- [x] `pnpm build` clean
- [x] Manual browser smoke on local stack (Playwright + local Chorus MCP):
  1. Create fresh session → visible in Settings (count=1, status=活跃)
  2. `UPDATE lastActiveAt = NOW() - 2h` → session hidden, new empty-state i18n key renders
  3. `chorus_session_heartbeat` → session reappears
- [x] OpenSpec validate passes; archive completed and mirrored back to Chorus document

## Notes for reviewers
- Migration is **DDL-only** (no `UPDATE` to backfill `inactive` rows). Per project policy, residual `inactive` rows from prior versions are handled by exclusion: the `status='active'` filter naturally excludes them, and the `heartbeatSession` recovery branch that previously revived them is removed.
- `touchLastActiveAt` includes a guard that skips the refresh when `status='closed'`, preserving the historical timestamp on closed sessions (raised by proposal-reviewer, implemented during execution, pinned by a unit test).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.9.0] - Unreleased
+
+### Changed (Breaking)
+- **Session lifecycle simplified to `{active, closed}`**: dropped the `inactive` status and the unused `expiresAt` field on `AgentSession`. "Stale" is now a query-time predicate on `lastActiveAt` (cutoff: 1h). Default UI listings (Settings page per-agent sessions, project worker-avatar header) only show active+fresh sessions; MCP-facing reads (`chorus_list_sessions`, `chorus_get_session`) and Activity-stream lookups remain unfiltered so plugin reuse and audit-trail navigation continue to see everything. Every session-touching MCP tool now refreshes `lastActiveAt` as a side effect of success — plugins no longer need standalone heartbeats during normal operation. The standalone `chorus_session_heartbeat` is retained for explicit keep-alives. The `expiresAt` parameter on `chorus_create_session` is removed.
+
+---
+
 ## [0.8.2] - 2026-05-21
 
 ### Added
--- a/docs/MCP_TOOLS.md
+++ b/docs/MCP_TOOLS.md
@@ -779,6 +779,14 @@ Supports intra-batch dependencies via `draftUuid` + `dependsOnDraftUuids`, and d
 
 Available to all Agents. Used to manage Agent work sessions (e.g., sub-agent workers in swarm mode).
 
+### Lifecycle and staleness
+
+The persisted state space is exactly `{active, closed}`. There is no `inactive` status — staleness is a derived predicate on `lastActiveAt`, not a stored value. Default UI listings (Settings page per-agent sessions, project worker-avatar header) only show sessions matching `status='active' AND lastActiveAt > now - 1h`. MCP-facing reads (`chorus_list_sessions`, `chorus_get_session`) and the audit-trail dereferences (Activity stream) are NOT filtered — plugins and history navigation see every session regardless of age or status.
+
+### Implicit-heartbeat contract
+
+Every Session tool that takes a `sessionUuid` parameter and successfully resolves the session refreshes `lastActiveAt = now()` as a side effect of success — except when the session's `status='closed'`, in which case the refresh is skipped to preserve the historical timestamp. Tools covered: `chorus_get_session`, `chorus_close_session`, `chorus_reopen_session`, `chorus_session_checkin_task`, `chorus_session_checkout_task`, `chorus_session_heartbeat`. As a result, plugins do not need to send standalone heartbeats during normal operation — any other session tool call already counts as one. The `chorus_session_heartbeat` tool remains as an explicit keep-alive for idle sub-agents.
+
 ### chorus_create_session
 
 **Description**: Create a new Agent Session (e.g., representing a sub-agent worker)
@@ -788,7 +796,6 @@ Available to all Agents. Used to manage Agent work sessions (e.g., sub-agent wor
 |-----------|------|----------|-------------|
 | name | string | Yes | Session name (e.g., "frontend-worker") |
 | description | string | No | Session description |
-| expiresAt | string | No | Expiration time (ISO timestamp) |
 
 **Output**:
 ```json
@@ -799,7 +806,6 @@ Available to all Agents. Used to manage Agent work sessions (e.g., sub-agent wor
   "description": "...",
   "status": "active",
   "lastActiveAt": "ISO timestamp",
-  "expiresAt": null,
   "createdAt": "ISO timestamp",
   "activeCheckins": []
 }
@@ -807,12 +813,12 @@ Available to all Agents. Used to manage Agent work sessions (e.g., sub-agent wor
 
 ### chorus_list_sessions
 
-**Description**: List all Sessions for the current Agent
+**Description**: List all Sessions for the current Agent. **Unfiltered** — returns sessions regardless of staleness or closed status (use this for plugin reuse logic and audit-trail navigation).
 
 **Input**:
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
-| status | string | No | Filter by status: active, inactive, closed |
+| status | string | No | Filter by status: `active` or `closed` |
 
 **Output**:
 ```json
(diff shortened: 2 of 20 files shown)


ChenNima (commented):
Solid surface refactor — the active/closed split is well-pinned by spec scenarios, the UI-vs-MCP read-path matrix is documented in the design doc, and the `inactive` regression-guard test is a nice touch. A few items worth addressing before merge:

### 1. Stale skill doc — `public/chorus-plugin/skills/develop/SKILL.md:363` still references `inactive`
The Impact section of the proposal explicitly lists `public/chorus-plugin/skills/chorus/` as needing the `inactive`/`expiresAt` cleanup, but the develop skill (same plugin tree) was missed:

```
| Session shows "inactive" (yellow) | No heartbeat in 1h. TeammateIdle hook should auto-send. Agent may have crossed |
```

Recommend rewording to match the new model — sessions never enter `inactive`; staleness is now the row dropping out of the Settings list after 1h. This was the only stale reference I found in `public/skill/` and `public/chorus-plugin/skills/`, so a single edit closes it out.

### 2. `reopenSession` lost its single-statement atomicity
Previously the function was one `prisma.agentSession.update` that flipped status and refreshed `lastActiveAt` together. After the refactor (`session.service.ts:387-403`) it's three separate statements:
1. `update { status: 'active' }`
2. `touchLastActiveAt` (findUnique + update lastActiveAt)
3. `findUniqueOrThrow` (re-read for response shape)

A concurrent `closeSession` landing between (1) and (3) would let `reopenSession` return a row with `status: 'closed'` while pretending it succeeded. Probability is low (humans clicking close + reopen on the same session simultaneously is rare), but the previous code couldn't hit it. Wrapping all three in `prisma.$transaction([...])` restores the invariant; alternatively, drop step (3) and do step (1) with a `select`/`include` so the response data and the status flip are one statement again — `touchLastActiveAt` remains a separate write but its lateness is then harmless because the response no longer depends on it.

### 3. `getSession` returns pre-touch `lastActiveAt`
`getSession` reads with `findFirst` then calls `touchLastActiveAt` after, returning the snapshot from the original read. So a UI that calls `getSession` and renders `lastActiveAt` will display the *previous* timestamp, then the *next* fetch will show the new one. Spec scenario "subsequent independent fetch MUST report `lastActiveAt >= T1`" passes (and the test exercises it that way), but a caller doing `getSession(...).lastActiveAt` immediately after will see the off-by-one. Cheap fix: do the touch first, then re-fetch (or reorder so the touch's update is the read source). Not blocking, but worth a one-liner comment in the function so a future reader doesn't get confused.

### 4. Migration leaves residual `inactive` rows untouched (intentional, well-documented — leaving a note for visibility)
Spec Requirement 1 explicitly accepts this and the design.md captures the rationale (DDL-only policy + filter excludes them). The risk note in the proposal flags it. Just confirming this is the intended trade-off, not an oversight — if there ARE non-trivial `inactive` rows in production, they'll keep matching `chorus_list_sessions` (unfiltered) responses indefinitely. The plugin's reuse logic queries by name+agent, so this likely doesn't surface as a user-visible bug, but worth a quick audit query post-deploy.

### 5. Nit: long-term capability spec has placeholder Purpose
`openspec/specs/session-lifecycle/spec.md:582` reads `TBD - created by archiving change session-lifecycle-active-only. Update Purpose after archive.` — that's the standard archive output but the TODO is now landing on `develop` along with the change. Worth filling in (one paragraph) before merge or as a fast-follow.

---

Items 1 and 2 I'd treat as merge-blockers; 3 is a nice-to-have; 4 and 5 are visibility / cleanup notes.

ChenNima (commented):
Fixup looks good — all four addressable items from the prior review are resolved cleanly:

- **`reopenSession` race**: collapsed to a single atomic `update` carrying both `status` and `lastActiveAt`. The new test pinning `update.calls.length === 1` and `findUnique.calls.length === 0` is exactly the right assertion shape.
- **`getSession` off-by-one timestamp**: snapshot mutation (`session.lastActiveAt = fresh`) before return, closed-row branch skips the touch. New test pair covers both directions (active row carries new timestamp; closed row keeps original).
- **Stale develop skill doc**: rewritten troubleshooting row matches the new "disappears from default lists after 1h" model and points readers at the heartbeat recovery path.
- **Capability spec Purpose**: real description covering all five sub-contracts (state machine, staleness filter, unfiltered audit reads, implicit heartbeat, task-status decoupling).

Minor note (not blocking): `getSession` now inlines the post-touch update logic instead of calling `touchLastActiveAt`, so there are two divergent code paths for "refresh lastActiveAt unless closed". The duplication is justified here (the helper does an extra findUnique that `getSession` already has the data for), and the comment captures the intent. Worth keeping an eye on if a third caller ever needs the same "snapshot + reflect touch" shape — would be the moment to factor out a `touchLastActiveAtAndReturn(session)` helper.

CI green. LGTM from my side — happy to see this merge.

PR #319, opened 2026-06-15
feat(daemon): self-reported SSE connection metadata + server-side registry

ChenNima:
## Summary

Closes idea **c152dcfc** ("SSE 连接自报元数据"). Makes long-lived daemon SSE connections **self-report** client metadata at connect time, and persists them server-side in a new `DaemonConnection` registry with reliable online/offline tracking.

This is deliberately the **data-collection + liveness layer only**. The read API, the Daemons observability UI, the per-connection session view, and the `AgentSession` linkage are all left to the derived idea **f2fe9a7f** (which this unblocks).

## How it works

- **Self-report contract** — clients append `?clientType=…&clientVersion=…&host=…&startedAt=…` to the SSE URL. Query params (not headers) so a future browser `EventSource`, which can't set headers, can use the same path. These params are display-only and **never** participate in auth (auth stays Bearer/cookie).
- **DB-backed registry** — a `DaemonConnection` row per logical daemon. DB (not in-memory/Redis) because prod runs 2 ECS tasks behind an ALB: a connection pinned to one instance must be visible to a query landing on the other. Reconnects upsert on `(agentUuid, clientType, host)`.
- **Liveness** — `abort` is the primary signal (graceful disconnect → `offline` immediately). The SSE route's existing 30s heartbeat interval also bumps `lastSeenAt`; since that interval dies with the instance process, it doubles as the instance-crash safety net (`status=online && lastSeenAt` fresh within ~90s = effectively online). **No client heartbeat added.**
- **Two clients self-report** — chorus CLI → `clientType=claude_code`, OpenClaw plugin → `clientType=openclaw` (versions read from each package.json).

## Backward compatibility

An old client or a browser that connects with **no `clientType`** (or an unknown one, or malformed params) connects **exactly as before** and writes no registry row. `parseSelfReport` never throws on absent/garbage input; `registerConnection` returns null before any DB write for non-daemon client types.

## Changed files

| File | Change |
|------|--------|
| `prisma/schema.prisma` + migration | New `DaemonConnection` model (DDL-only migration via Prisma CLI) |
| `src/services/daemon-connection.service.ts` (+ test) | register / markDisconnected / touchConnection, companyUuid-scoped, fire-and-forget |
| `src/app/api/events/route.ts`, `…/notifications/route.ts` (+ tests) | register on connect, offline on abort, touch on heartbeat |
| `cli/sse-listener.mjs` (+ test) | self-report `claude_code` |
| `packages/openclaw-plugin/src/sse-listener.ts` (+ test) | self-report `openclaw` |
| `openspec/specs/*`, `openspec/changes/archive/*` | archived OpenSpec change + cumulative specs |

## Test plan

- [x] `daemon-connection.service` unit tests (Prisma mocked): register/upsert/disconnect/touch/swallowed-error — 100% coverage on the new service
- [x] Both SSE route tests: daemon register-on-connect + touch-on-heartbeat + offline-on-abort; no-clientType connection still streams, writes nothing
- [x] Both client tests: URL self-report params + version-from-package.json + reconnect re-sends
- [x] Live E2E: real chorus CLI daemon → row appears online → `lastSeenAt` advances on tick → graceful shutdown → offline + `disconnectedAt` → reconnect refreshes same row → parallel openclaw → distinct row
- [x] Full suite green (2031 passed / 1 skipped); `tsc --noEmit` clean

### DIFF
--- a/cli/__tests__/sse-listener.test.mjs
+++ b/cli/__tests__/sse-listener.test.mjs
@@ -1,8 +1,20 @@
 // cli/__tests__/sse-listener.test.mjs
 // Covers cli-daemon spec "Daemon subcommand and notification subscription"
-// (SSE parsing) and the backoff/reconnect behavior.
+// (SSE parsing), the backoff/reconnect behavior, and the self-report query
+// params the listener appends to the notification SSE URL.
 import { describe, it, expect, vi } from "vitest";
+import { readFileSync } from "node:fs";
+import { hostname } from "node:os";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
 import { SseListener } from "../sse-listener.mjs";
+
+// The version the listener self-reports comes from the chorus CLI's own
+// package.json (one level above cli/) — assert against that same source rather
+// than a hardcoded literal, so a version bump doesn't break the test.
+const CLI_VERSION = JSON.parse(
+  readFileSync(join(dirname(fileURLToPath(import.meta.url)), "..", "..", "package.json"), "utf8")
+).version;
 
 /** Build a fetch Response whose body streams the given chunks then ends. */
 function streamingResponse(chunks, { ok = true, status = 200 } = {}) {
@@ -45,11 +57,13 @@ describe("SseListener parsing", () => {
       { type: "new_notification", notificationUuid: "n1" },
       { type: "count_update", unreadCount: 3 },
     ]);
-    // Sent Bearer auth to the right endpoint
-    expect(fetchImpl).toHaveBeenCalledWith(
-      "https://chorus.example/api/events/notifications",
-      expect.objectContaining({ headers: expect.objectContaining({ Authorization: "Bearer cho_x" }) })
+    // Sent Bearer auth to the notification endpoint (now carrying self-report
+    // query params — assert the path + Bearer header, params covered below).
+    const [calledUrl, calledInit] = fetchImpl.mock.calls[0];
+    expect(calledUrl).toMatch(
+      /^https:\/\/chorus\.example\/api\/events\/notifications\?/
     );
+    expect(calledInit.headers).toMatchObject({ Authorization: "Bearer cho_x" });
   });
 
   it("strips trailing CR so CRLF transports parse", async () => {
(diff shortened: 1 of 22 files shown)


ChenNima:
## Review — overall: high quality, good to merge ✅

Strong test coverage (service 100%, both SSE routes, both clients), correct multi-tenancy, solid backward compat, and faithful to project conventions (Prisma-CLI DDL-only migration, service layer, fire-and-forget liveness, no UI so no i18n/design.pen needed). CI `test` is green. No blockers — the points below are NOTE-level.

### 1. Cross-connection reconnect race (the only real correctness note · low-severity · self-healing)

`markDisconnected` and `registerConnection` both act on the **same row** (deterministic key `(agentUuid, clientType, host)`). Consider:

- Old connection (interval/abort listener holding `connUuid=X`) drops
- Client backs off and reconnects (≥1s) → `registerConnection` upserts row X back to `online`
- **Then** the old connection's `abort` finally fires → `markDisconnected(X)` writes the now-online row to `offline`

The common path is safe (socket close → abort fires near-immediately → offline, *then* register online 1s+ later). But if the old socket's close detection lags behind the client's backoff (half-open socket, delayed fetch abort), you get an "online-then-stale-abort-flips-to-offline" inversion. This can hit single-instance too, not just the 2-ECS case.

Impact is bounded: the next 30s `touchConnection` re-asserts `status=online`, so the worst case is ≤30s of wrong `offline`; and it's display-only with no consumer yet (`f2fe9a7f`). Hence NOTE, not blocker.

Optional clean fix if you want to kill it now: have `registerConnection` also return `connectedAt`, and make `markDisconnected` a conditional update `where: { uuid, companyUuid, connectedAt: <captured> }`. A reconnect refreshes `connectedAt`, so the stale abort's conditional update matches 0 rows and the race disappears. Low cost.

### 2. `design.md` says "all registry calls are fire-and-forget", but `registerConnection` is `await`ed on the connect path

`const connUuid = await registerConnection(...)` (`route.ts:29/31`) is necessary to obtain `connUuid` and is fine. It does add one synchronous DB round-trip to daemon connection setup (browsers are unaffected — a non-daemon clientType returns `null` before any DB call, which is implemented cleanly). Just the design doc wording ("All registry calls are fire-and-forget (`void` + internal try/catch)") doesn't match: register is "awaited but non-throwing". Worth a one-line clarification.

### 3. Upsert key doesn't include the SSE endpoint

`(agentUuid, clientType, host)` doesn't distinguish `/api/events` from `/api/events/notifications`. Fine today (CLI/OpenClaw only hit notifications; browser hits events with no clientType so isn't registered). But **if browser registration is added later** (on `/api/events`) and a clientType ever overlaps a daemon's, two connections would upsert onto the same row and fight. Worth keeping in mind for `f2fe9a7f`.

### 4. Minor doc/schema drift (non-blocking)

- Archived `design.md` / `proposal.md` say `host String?` (nullable); final schema is `host String @default("")` (non-null). The **final decision is the better one** (a `@@unique` over a nullable column treats each null as distinct and breaks dedup — the schema comment explains this well); the archived design doc just wasn't synced.
- Cumulative spec `openspec/specs/daemon-connection-registry/spec.md` still has the `Purpose: TBD - created by archiving` placeholder at the top — easy to fill in.

### Nicely done
- Defensive `parseSelfReport` for `startedAt` (empty/invalid → `null`, never an `Invalid Date`), with dedicated tests.
- Versions read from each package.json with a `"0.0.0"` fallback; the relative paths resolve to the right manifest across `src`/`dist`/jiti-linked forms (verified), and tests assert against the same source so a version bump won't break them.
- Strict multi-tenancy: `companyUuid`/`actorUuid` always from the auth context, query params never authorize — and tests assert exactly that.
- 401 returns *before* `parseSelfReport`, so unauthenticated connections incur zero extra cost.

🤖 Review by Claude Code

ChenNima:
Thanks for the thorough review. Addressed in `9e33db3`:

**#1 Reconnect race (the one real correctness note) — fixed.** Took a slightly stronger version of your suggestion: `registerConnection` now returns a `{ uuid, connectedAt }` handle, and **both** `touchConnection` and `markDisconnected` gate their `updateMany` on `connectedAt: handle.connectedAt`. Each (re)registration stamps a fresh `connectedAt`, so any call from an older connection generation matches 0 rows and is a no-op. This kills the whole cross-generation class — not just the stale-`abort`→offline inversion you described, but also a lingering old heartbeat keeping a re-registered row alive. The worst prior case (≤30s of wrong `offline`, self-healing) is now structurally impossible. Service file stays at 100% coverage; +4 fence no-op tests assert the count-0 path neither throws nor logs.

**#4 Spec Purpose placeholder — fixed.** `openspec/specs/daemon-connection-registry/spec.md` had the `Purpose: TBD - created by archiving` boilerplate; filled it in and mirrored back to the Chorus doc.

**#2 / #3 — intentionally not changed, noted for the record:**
- **#2** (`design.md` says "fire-and-forget" but `registerConnection` is awaited): that wording lives in the **archived** change doc (`openspec/changes/archive/…`), which is a point-in-time snapshot of what was proposed. Leaving it as history rather than rewriting an archived artifact. The awaited-but-non-throwing register is correct and unchanged.
- **#3** (upsert key omits the SSE endpoint): a genuine concern only if browser registration later lands on `/api/events` with a clientType that overlaps a daemon's. That's squarely **f2fe9a7f** territory and is already captured as a boundary note for that idea. No change here keeps this PR scoped to collection+liveness.

CI was green before; re-running on the new commit.

PR #323, opened 2026-06-16
feat(daemon): execution-state reporting + UI (running/queued per connection)

ChenNima:
## Summary

Surfaces the daemon's **running/queued tasks** on the Agent Connections detail pane (replacing the prior `ComingSoonPlaceholder`), in real time. Implements sub-idea 子2 (`3b7711b3`) of the daemon dispatch umbrella (`6fab91cd`).

- New persistent **`DaemonTaskExecution`** model (one row per connection/task; snapshot-reconciled `running`/`queued`/`ended`).
- New agent-key REST ingest endpoint **`POST /api/daemon/execution-state`** (+ `GET` first-paint read), fed from the daemon's `WakeQueue`. Not an MCP tool, no new permission bit.
- New additive **`execution:{connectionUuid}` SSE event** over the existing EventBus (+ Redis fan-out) — deliberately diverges from the connection page's polling because execution changes far faster than liveness.
- **Owner-scoped visibility** identical to the connection registry. An offline/stale connection shows **no active execution** (read-time staleness gate reusing the registry's single `STALE_THRESHOLD_MS`), while rows are retained as queryable history.
- Daemon side (npm CLI) uploads its snapshot fire-and-forget from real WakeQueue state; zero new dependencies.

## Changed files
| Area | Files |
|------|-------|
| Schema | `prisma/schema.prisma`, `prisma/migrations/20260616152622_add_daemon_task_execution/` (DDL-only) |
| Service | `src/services/daemon-execution.service.ts` (+ tests) |
| API | `src/app/api/daemon/execution-state/route.ts` (+ tests); `src/app/api/events/route.ts`, `.../events/notifications/route.ts` (SSE forward + offline reconcile) |
| Realtime | `src/contexts/realtime-context.tsx` (`subscribeExecution` / `useExecutionSubscription`) |
| UI | `src/app/(dashboard)/agent-connections/page.tsx` (ExecutionPane) (+ test); `messages/{en,zh}.json` |
| Daemon CLI | `cli/{upload-hooks,waker,wake-queue,event-router,sse-listener,daemon}.mjs` (+ tests) |
| Spec | `openspec/specs/daemon-execution-state/`, archived change under `openspec/changes/archive/` |

## Test plan
- [x] `npx tsc --noEmit` → clean
- [x] Touched suites green (110 tests: service 48, ingest route 13, events route 14, notifications 7, execution-view UI 5, CLI upload-hooks 13 + sse-listener 10)
- [x] Full suite during development: 2146 passed / 0 failed
- [x] Zero new npm dependencies (`git diff develop -- package.json` empty)
- [x] Independent code review: **APPROVE WITH NOTES**, 0 blockers, 0 security issues
- [x] Real-server chain verified: agent-key POST → reconcile → SSE → browser render; offline → empty + history retained

## Notes / follow-ups (non-blocking)
- `docs/design.pen` not updated — Pencil MCP transport unavailable in the build env; tracked as a follow-up.
- Late-connecting daemon can lag ~15s on an already-open page (poll-mitigated; documented in the SSE route).
- Pre-existing: EventBus has no `setMaxListeners` (>10 concurrent SSE streams warn) — separate follow-up, not introduced here.
- Siblings 子1 (concurrency cap) and 子3 (interrupt/resume) remain.


ChenNima:
## Follow-up fix: report execution for every wake resource (was task-only) + RealtimeProvider wiring

Live testing surfaced two real bugs in the just-shipped feature; this commit (`c54681f`) fixes both.

**Bug 1 — empty UI (root cause).** The daemon only recorded a `DaemonExecution` row when the wake notification's `entityType === "task"`. @mentioning an agent **under an idea** woke and ran Claude but produced **zero rows** → the execution pane was empty. Generalized to report **any** wake-triggering resource:
- Model: `DaemonTaskExecution` → `DaemonExecution`; `taskUuid` → `entityType` + `entityUuid` (`task | idea | proposal | document`); unique `(connectionUuid, entityType, entityUuid)`.
- Migration done the right way: the already-applied `20260616152622_add_daemon_task_execution` is **preserved**, and a new Prisma-CLI migration `20260617013334_generalize_daemon_execution_to_any_resource` transforms it.
- `cli/waker.mjs` records every recognized wake across queued→running→finish; route/zod/service generalized; service enriches title/project/link per entity kind.

**Bug 2 — live updates never fired (the earlier review's 🔴 BLOCKER).** `/agent-connections` wasn't wrapped in `RealtimeProvider` (the dashboard layout only mounts one for project pages), so `useExecutionSubscription` silently no-op'd. Fixed by wrapping the page's default export in `<RealtimeProvider>`; the execution-view test now renders the page without a test-supplied provider so it exercises the real wiring.

**Also folded in (review Medium-1):** entity validation is now best-effort (`filterValidExecutionEntities`) — a dead reference is dropped and the rest reconcile, instead of 400-rejecting the whole snapshot.

**Verification**
- `npx tsc --noEmit` clean; full suite **2154 passed / 0 failed**.
- **Live against the running server**: idea-wake snapshot POST → `reconciled: 2` → GET read-back returns the running idea + queued task, enriched (title/project/session). Foreign-company ref → dropped. Test rows cleaned up.
- Independent task-reviewer: **PASS WITH NOTES** (confirmed the migration was done correctly; 2 stale comments since fixed).

**Follow-up (not in this commit):** the archived cumulative spec `openspec/specs/daemon-execution-state/spec.md` still describes the old task-only model. Correcting it should be a new OpenSpec change with `## MODIFIED Requirements`, not a hand-edit of the archived spec — flagging rather than doing it unilaterally.

PR #394, opened 2026-07-04
feat(daemon): server-side orphan-turn reconcile for dead daemons

ChenNima:
## Summary

Task 2 of 5 for idea 489ade18 — stacked on #393 (base: feat/turn-interrupted-state).

- New `reconcileOrphanTurns(companyUuid, connectionUuid)`: finalizes `running` turns of the connection's sessions (owner = `session.originConnectionUuid`, the origin-pinned anchor) as `interrupted(offline)` through the `advanceTurn` chokepoint — legality check + `turn_status_changed` SSE come free. `pending` turns untouched (backfill's job). Fire-and-forget-safe (swallow+log, mirrors execution `reconcileOffline`).
- **Age-only orphan eligibility** (round-1 review blocker): a connection is orphan-eligible only when `lastSeenAt` is older than the registry's single `STALE_THRESHOLD_MS` (90s) — never the stored `status` alone, which flips on every transient SSE abort. The eligibility read selects only `lastSeenAt`, so status cannot influence the verdict.
- **Trigger 1 — deferred abort:** both SSE routes (`/api/events`, `/api/events/notifications`) arm an unref'd `setTimeout(STALE_THRESHOLD_MS)` on daemon stream abort; the age re-check at fire time makes it a no-op when the daemon reconnected.
- **Trigger 2 — read-time fallback:** `getVisibleSessions` + `getSessionDetail` converge orphaned running turns write-through before building views — this also heals legacy dirty rows and the lost-timer (server restart) case with zero migration DML. The fallback is best-effort end-to-end (a probe failure is logged, the read continues).
- Race with the daemon's own terminal report: loser gets a logged `invalid_transition`, writes nothing.

## Testing

- New suites: `reconcileOrphanTurns` (stale verdict; fresh-lastSeenAt no-op **regardless of status** — abort→reconnect-gap regression test; kill-9 eligibility while status still 'online'; deleted connection; pending untouched; race → logged invalid_transition; swallow+log) and read-time fallback (list + detail converge; gap read doesn't interrupt; read survives fallback failure).
- Both SSE route tests: deferred timer fires exactly at the window, not before; browser streams never arm it.
- `pnpm test`: 3903 passed (242 files). `npx tsc --noEmit` clean. `pnpm lint` 0 errors.

### DIFF
--- a/src/app/api/events/__tests__/route.test.ts
+++ b/src/app/api/events/__tests__/route.test.ts
@@ -19,6 +19,7 @@ const mockReconcileOffline = vi.fn();
 const mockPublishExecutionChange = vi.fn();
 const mockListVisibleConnectionUuids = vi.fn();
 const mockIsSessionVisibleToCaller = vi.fn();
+const mockReconcileOrphanTurns = vi.fn();
 
 vi.mock("@/lib/auth", () => ({
   getAuthContext: (...args: unknown[]) => mockGetAuthContext(...args),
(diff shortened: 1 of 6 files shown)


ChenNima:
Superseded — all commits from this branch landed on develop via the integration PR #397 (merge 44790a4). Closing without a direct merge.

PR #457, opened 2026-07-28
feat(pi): Chorus AI-DLC extension for the Pi coding agent + sync v0.14.5

expoli:
## Summary

Ports the Chorus AI-DLC plugin to the [Pi](https://pi.dev) coding agent and syncs it to upstream v0.14.5. Pi is the lowest-friction plugin target — MCP auto-discovery (no installer), TypeScript hooks (one file replaces ~10 bash scripts), and automatic sub-agent session lifecycle (a capability the Codex port lacks).

### 1. The `chorus-pi` package (new)

- **8 Agent Skills** — `/skill:chorus`, `/skill:idea`, `/skill:proposal`, `/skill:develop`, `/skill:review`, `/skill:quick-dev`, `/skill:yolo`, `/skill:openspec-aware`
- **3 read-only reviewer sub-agents** — `chorus-proposal-reviewer`, `chorus-task-reviewer`, `chorus-code-reviewer` (read-only: `read, grep, find, ls, bash, mcp` — no write/edit)
- **1 session-aware extension** (`extensions/chorus.ts`) — subscribes to Pi native events to automate checkin, context injection, reviewer nudges, and session lifecycle
- **OpenSpec wrapper** (`bin/chorus-mcp-call.sh`) — stateless MCP-over-HTTP for byte-exact document mirroring

### 2. Sync to upstream v0.14.5

The port was originally based on v0.14.3 content. After the port, upstream released v0.14.4 (daemon backend config) and v0.14.5 (#454: daemon now exports `CHORUS_URL` + `CHORUS_API_KEY` into every spawner's child env). Both are daemon/cli changes — `chorus-pi` already reads those env vars (`extensions/chorus.ts` + `bin/chorus-mcp-call.sh`), so it's compatible with zero changes and actually *benefits* from #454 (woken Pi sessions auto-receive the connection pair).

### 3. User-visible SessionStart banner (#442 parity)

The Claude/Codex plugins emit a one-line user-visible toast at SessionStart (`(OpenSpec Enabled)` / `(OpenSpec off)` / `(OpenSpec off — run … enable openspec …)`), but the Pi port only injected a hidden agent-context block (`display:false`) — the skills even referenced the `(OpenSpec off — …)` banner text without the extension ever producing it. Added `buildSessionBanner()` (pure, unit-tested) and emit it via `ctx.ui.notify` in `session_start`:

| State | Banner |
|-------|--------|
| connected + OpenSpec active | `Chorus connected at <url> (OpenSpec Enabled)` |
| connected + explicit opt-out | `Chorus connected at <url> (OpenSpec off)` [neutral, no nag] |
| connected + not set up | `Chorus connected at <url> (OpenSpec off — run /skill:chorus enable openspec to set it up)` |
| not configured (env missing) | warning: `Chorus plugin: not configured (set CHORUS_URL and CHORUS_API_KEY)` |
| connection failed | error: `Chorus: connection failed (<url>)` |

### 4. `maxCodeReviewRounds` as `CHORUS_MAX_CODE_REVIEW_ROUNDS` env

The port had dropped the Claude plugin's `maxCodeReviewRounds` userConfig (default 3, 0 = unlimited) from the chorus skill config table (Pi has no plugin settings UI → env vars drive the toggles), but the yolo/develop/review skills still referenced `maxCodeReviewRounds` as if it were a plugin config — a documentation inconsistency leaving the agent with no resolvable value for the code-review-gateway loop cap. Restored it as the env var `CHORUS_MAX_CODE_REVIEW_ROUNDS`:

- `parseMaxCodeReviewRounds()` pure helper in `lib/lib.ts` (default 3, 0 = unlimited; invalid/blank/non-integer/negative all fall back to default instead of silently flooring)
- Extension reads the env once at load and surfaces the resolved value in the injected Quick Reference so the agent knows the loop cap without re-reading the env
- chorus skill §5 config table: row + `export …=5` example
- yolo/develop/review skills: every `maxCodeReviewRounds` reference → `CHORUS_MAX_CODE_REVIEW_ROUNDS`

## Why Pi

- **MCP: zero installer.** `pi-mcp-adapter` auto-discovers the repo's `.mcp.json` (literal URL + Bearer — no `${VAR}` expansion). Main agent gets all 40+ `chorus_*` tools with no setup script.
- **Hooks: TypeScript, not bash.** One TS file replaces ~10 bash hook scripts. No `curl`/`jq`, no Bash 3.2 compatibility traps (the `${2:-{}}` JSON-parse bug that plagued the Codex port is structurally impossible here).
- **Sub-agent sessions: automatic.** By monitoring `subagent_spawn` / `subagent_manage` tool events, the extension auto-creates and closes Chorus sessions — a capability the Codex port lacks (Codex workers manage sessions manually).
- **Skills: same standard.** Pi implements the Agent Skills standard, so the skill bodies port with find/replace only (Claude's `Task` tool → `subagent_spawn`; `/chorus:develop` → `/skill:develop`).

## Verification

- **Layer A static**: 32 passed — TS transpile, package.json/manifest valid, skill+agent frontmatter compliant, no Claude/Codex product residual, `/skill:X` cross-refs all resolve, `agent:` spawns all exist
- **Layer B unit**: 42 passed (37 existing + 5 new parser tests + 7 banner tests)
- `chorus.ts` transpiles clean; no module-scope undefined vars

## Files

```
chorus-pi/
├── package.json              # pi.extension entry + peerDeps (v0.14.5)
├── extensions/chorus.ts      # session_start / before_agent_start / tool_call / tool_result / tool_execution_end / session_shutdown
├── lib/lib.ts                 # pure helpers (detectOpenSpec, buildSessionBanner, parseMaxCodeReviewRounds, …)
├── skills/                    # 8 Agent Skills standard SKILL.md
├── agents/                    # 3 reviewer sub-agents
├── bin/chorus-mcp-call.sh     # stateless MCP-over-HTTP wrapper (OpenSpec mirroring)
└── test/                      # all.sh (A+B), static.sh, lib.test.ts, precheck.sh, verify-pi-session.md
```

Follow-up `docs/CONNECT_PI.md` covers install + env config.


ChenNima:
## Code review findings

I reviewed PR #457 at head `4f32c83543a467807603c287209477117eead111` against base `be647877b4b56a61e480e939d6a6d31b3f84f7f9`.

### [P1] Continue past incomplete `.mcp.json` candidates

`chorus-pi/lib/lib.ts:74-78` returns the first candidate containing either a URL or an API key, and `chorus-pi/bin/chorus-mcp-call.sh:31-49` similarly stops at the first `mcpServers.chorus` entry. If the project config contains only one field while `~/.pi/agent/mcp.json` contains a complete configuration, the usable global candidate is never considered. The extension is then treated as unconfigured and the wrapper exits.

I reproduced this against the PR head: a project candidate with only `url` resolves to `{ url: "https://project/api/mcp", apiKey: "" }` instead of falling through to the complete global candidate. Please accept a candidate only when the required fields are complete, or continue searching after field-level env/config merging remains incomplete. Both TS and shell paths need a partial-candidate regression test.

### [P1] Close the Chorus session when agent ID extraction fails

In `chorus-pi/extensions/chorus.ts:369-391`, `pendingSessions.delete(event.toolCallId)` runs before the fallback extraction result is known. If a successful `subagent_spawn` returns a shape neither extractor recognizes, the session is no longer in `pendingSessions`, is never added to `sessionMap`, and is not closed. Neither `subagent_manage close` nor `session_shutdown` can recover it, so it remains active indefinitely.

Please close the session immediately on fallback extraction failure, or retain it in an orphan collection that shutdown can retry. This branch also needs an extension-event test using an unknown spawn result shape.

### [P1] Do not delete the session mapping after a failed close

At `chorus-pi/extensions/chorus.ts:321-325`, `chorus_close_session` errors are swallowed, but the mapping is deleted and the UI reports success regardless. A transient network/server failure therefore becomes a permanent leaked backend session, and shutdown cannot retry it.

Please remove the mapping and emit the success notification only after a successful close. On failure, retain it for shutdown retry and report a warning.

### [P2] Restrict task-session injection to actual workers

`chorus-pi/extensions/chorus.ts:271-282` excludes only the three Chorus reviewer names. Despite the comment saying non-worker agents do not get a Chorus session, built-in `scout`, `planner`, `reviewer`, and arbitrary custom agents all receive a session plus task check-in/update/report instructions.

This adds unnecessary API traffic and injects irrelevant task lifecycle instructions into read-only agents. Please use an explicit worker allowlist (at minimum `agentName === "worker"`) or another positive worker classification.

### [P2] Include reviewer-agent installation in the package README flow

The primary install instructions in `chorus-pi/README.md:11-24` stop after installing dependencies and the package, then direct users only to MCP/env configuration. However, `chorus-pi/package.json:29-33` loads only extensions and skills, and there is no installer/postinstall that copies `agents/*.md`. Users following the package README therefore cannot spawn any of the mandatory proposal/task/code reviewers, and YOLO fails at its first review gate.

`docs/CONNECT_PI.md:65-74` contains the missing manual copy step. Please add it to the package README install flow or provide a reliable installer/postinstall path.

### Verification

- `bash test/all.sh`: passed (`32` static checks, `54` unit tests)
- `git diff --check`: two trailing-whitespace errors in `chorus-pi/skills/yolo/SKILL.md:312,441`
- `bash test/precheck.sh`: `4` passed / `4` failed because this host lacks the Pi runtime dependencies, Pi MCP config, and reviewer discovery files; fresh-Pi C/D/E integration layers were not run
- GitHub currently exposes no status checks for this PR

ChenNima:
## Re-review of `46110ff0`

I re-reviewed the update from `4f32c835` to `46110ff0` and reran the updated offline suite.

Four findings are resolved:

- incomplete `.mcp.json` candidates now fall through to a complete candidate, with regression tests;
- failed mapped-session closes retain the mapping for shutdown retry;
- session injection is positively gated to the canonical worker;
- the package README now includes reviewer-agent installation.

The trailing whitespace is also fixed. `git diff --check` passes, and `bash test/all.sh` passes with `32` static checks, `62` helper tests, and `4` extension-event tests.

### [P1] Retain orphan sessions when the immediate close fails

The agent-ID extraction fix at `chorus-pi/extensions/chorus.ts:403-414` still performs:

```ts
await mcpCall("chorus_close_session", { sessionUuid: sid }).catch(() => {});
```

After this call, the `sid` is not retained anywhere and the warning always says the orphan session was closed. If the close request fails due to a transient network/server error, the original leak remains: `pendingSessions` was already deleted, the session is absent from `sessionMap`, and `session_shutdown` cannot retry it.

The new `P1-2` event test only asserts that a close call was attempted; it does not cover close failure or verify a shutdown retry.

Please retain the orphan session in a retryable collection (or restore it to `pendingSessions`) when close fails, remove it only after success, and report success/failure accurately. Add a test where the orphan close returns an MCP error and assert that `session_shutdown` retries it. The same concern applies to the spawn-error close at `chorus-pi/extensions/chorus.ts:389-391`, which also swallows failure after deleting the pending entry.

PR #463, opened 2026-07-30
feat(mcp): bound collection responses

ChenNima:
## Summary
- standardize bounded pagination and compact projections for MCP collection tools
- enforce a 65,536-byte serialized response ceiling for collections and aggregates
- add exact UUID search plus exhaustive collection-tool classification and inventory checks
- synchronize MCP usage guidance across maintained plugin surfaces

## Verification
- [x] 1,981 MCP/service tests passed
- [x] TypeScript and production build passed
- [x] Deployed with `./default_deploy.sh`
- [x] Production Streamable HTTP MCP E2E: 18/18 collection and aggregate tools, 33 calls
- [x] Default/max pagination, metadata, compact projection, UUID search/get all passed
- [x] Largest production response: 37,594 bytes; zero mutations and no data residue

## OpenSpec
- archived change: `bound-mcp-list-responses`
- cumulative spec: `openspec/specs/bounded-mcp-collections/spec.md`

### DIFF
--- a/docs/MCP_TOOLS.md
+++ b/docs/MCP_TOOLS.md
@@ -380,14 +380,14 @@ Each idea entry carries the stored single-parent lineage edge as `parentUuid` (o
 
 ### chorus_get_activity
 
-**Description**: Get the activity stream for a project
+**Description**: Get the paginated activity stream for a project. There is no single-activity get tool, so each row retains the full activity payload, including `value` and session context.
 
 **Input**:
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
 | projectUuid | string | Yes | Project UUID |
 | page | number | No | Page number |
-| pageSize | number | No | Items per page (default 50) |
+| pageSize | number | No | Items per page (default 20, maximum 100) |
 
 **Output**: Activity list JSON
 
@@ -713,7 +713,7 @@ Each task in the response includes the full TaskResponse format (with dependsOn,
 
 ### chorus_get_comments
 
-**Description**: Get the list of comments for an Idea/Proposal/Task/Document
+**Description**: Get paginated comments for an Idea/Proposal/Task/Document, newest first. There is no single-comment get tool, so each row includes the full `content` and resolved `author`.
 
 **Input**:
 | Parameter | Type | Required | Description |
@@ -721,7 +721,7 @@ Each task in the response includes the full TaskResponse format (with dependsOn,
 | targetType | enum | Yes | Target type: idea, proposal, task, document |
 | targetUuid | string | Yes | Target UUID |
 | page | number | No | Page number |
-| pageSize | number | No | Items per page |
+| pageSize | number | No | Items per page (default 20, maximum 100) |
 
 **Output**: Comment list JSON
 
--- a/docs/MCP_TOOLS.md
+++ b/docs/MCP_TOOLS.md
@@ -389,7 +389,7 @@ Each idea entry carries the stored single-parent lineage edge as `parentUuid` (o
 | page | number | No | Page number |
 | pageSize | number | No | Items per page (default 20, maximum 100) |
 
-**Output**: Activity list JSON
+**Output**: Activity list JSON. All page-mode collection responses include `returned`, `page`, `pageSize`, and the full matching `total`. Bounded discovery tools may instead report `total` as the returned candidate-set size when the backing query does not expose a full count.
 
 ### chorus_get_my_assignments
 
--- a/docs/MCP_TOOLS.md
+++ b/docs/MCP_TOOLS.md
@@ -571,12 +571,12 @@ Each task in the response includes the full TaskResponse format (with dependsOn,
 
 ### chorus_search
 
-**Description**: Search across tasks, ideas, proposals, documents, projects, and project groups. Supports scoping to project groups or specific projects.
+**Description**: Search compact summaries across tasks, ideas, proposals, documents, projects, and project groups. A canonical UUID query performs tenant-scoped exact lookup first; text search is the fallback. Prefer search for discovery, use paginated list tools only for browsing, and call the entity's single-resource `get` tool for full details.
 
 **Input**:
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
-| query | string | Yes | Search query (matches title, description, content) |
+| query | string | Yes | Canonical entity UUID for exact lookup, or text matching title, description, and content |
 | scope | enum | No | Search scope: global, group, project (default: global) |
 | scopeUuid | string | No | Project group UUID (scope=group) or project UUID (scope=project) |
 | entityTypes | string[] | No | Entity types to search: task, idea, proposal, document, project, project_group (default: all) |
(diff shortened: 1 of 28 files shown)


ChenNima:
## Code Review — `feat(mcp): bound collection responses`

Reviewed the full diff (25 files, +2128/-104). Overall this is a well-structured, well-tested change: a single shared collection contract (`collection-contract.ts`), an executable inventory + registration-time enforcement, per-tool compact projections, and an exact-UUID search path. Ran the affected suites locally — **60/60 pass** (`collection-contract`, `collection-migration`, `server`, `search.service`). Findings below, most-substantive first.

### 1. `chorus_search` advertises `pageSize: 50` but can only ever return 20 rows (medium)
The handler calls `searchService.search({...})` **without** passing `limit`, so the service falls back to its default `limit = 20` and does `allResults.slice(0, 20)`. Yet the handler reports `pageSize: 50`, and `COLLECTION_TOOL_INVENTORY.chorus_search` documents `maximumRows: 50` with the reason *"Search service enforces a cross-entity result limit."* The real enforced limit is **20**, not 50 — the metadata and the inventory reason are both inaccurate, and no test pins the actual cap so it drifted silently. Either pass `limit: 50` into the service or correct `pageSize`/`maximumRows` to `20`.

### 2. `total` means different things on page tools vs. bounded tools (medium)
Page tools (`chorus_list_tasks`, etc.) report the true DB `total` (e.g. `500`), which correctly signals "there's more." The bounded tools — `chorus_search`, `chorus_search_mentionables`, `chorus_get_available_ideas/tasks`, `chorus_get_project_groups` — report `total = returnedArray.length`. So an agent that reads `total: 20` on a search will conclude only 20 matches exist corpus-wide. Consider either documenting that bounded-tool `total` == `returned` (not corpus size), or omitting `total` for those tools. (For search it's largely redundant with the `counts` object that's already carried in `extra`.)

### 3. `chorus_get_unblocked_tasks`: rows beyond 100 are unreachable (medium/low)
`getUnblockedTasks` has no server-side `take` — it returns **all** unblocked tasks plus a true `total`. The handler then does `tasks.slice(0, 100)` with **no `page`/`pageSize` input**. If a project has >100 unblocked tasks (or the 64 KiB cap trims further), `total` reads e.g. `150` while only 100 rows come back and there is no pagination parameter to fetch the remainder. Either give it page params like the page-mode tools, or document the hard cap explicitly. The same "no page param + silent cap" structure applies to `chorus_get_available_ideas/tasks` (service caps at `take: 50`, inventory says `maximumRows: 100`) and `chorus_get_project_groups` — lower risk since overflow is less likely, but the inventory `maximumRows` for available ideas/tasks doesn't match the real service cap of 50.

### 4. `serializeBoundedAggregate` can hard-fail `chorus_checkin` on long scalars (low)
The aggregate trimmer only shrinks arrays under the hardcoded keys `ideas/tasks/notifications/recentActivities` and never truncates scalar strings. For `chorus_checkin`, long scalar fields such as `agent.persona` / `agent.systemPrompt` are not trimmable; if they alone push the payload past 64 KiB, the tool throws `CollectionPayloadTooLargeError` — a hard error **on the session-start path** rather than a graceful degrade. Low probability, but consider truncating known long scalars (persona/systemPrompt) or catching and degrading rather than throwing at check-in.

### 5. `compactTracker` drops whole projects once the budget is spent (low)
Once the shared 100-item budget hits zero (`if (remaining === 0) break;`), the loop stops adding *projects* entirely, so later projects vanish from the tracker map (not just their items). Acceptable for a bounded aggregate, but worth a code comment so it isn't mistaken for a bug later.

### Nits / positives
- `buildBoundedCollectionPayload` re-serializes the entire payload on each `pop()`. Bounded by `pageSize ≤ 100` and 64 KiB, so fine in practice — just noting the O(rows) stringify-per-pop if it ever grows.
- Good backward-compat awareness: `on-subagent-start.sh` already handles both the old bare-array and the new `{ sessions: [...] }` shape for `chorus_list_sessions`.
- The registration-time `enforceToolClassification` Proxy + the "must equal registered tools" test is a strong guardrail against future collection tools slipping the contract.
- Truncation helpers correctly operate on Unicode code points (surrogate-pair safe), and the byte-ceiling logic is verified with genuine multibyte fixtures.

Nothing here is a blocker; #1–#3 are mostly metadata-accuracy issues that could mislead an agent's paging decisions, and #1 is a quick fix worth doing before merge.

ChenNima:
Addressed all review findings in `14cebf7`:

- `chorus_search` now passes `limit: 50`; `pageSize`, inventory cap, and actual results align, and `total` is derived from entity counts.
- `chorus_get_unblocked_tasks` and `chorus_get_project_groups` now support `page`/`pageSize` with reachable subsequent pages.
- available idea/task discovery caps are documented and inventoried at the actual 50-row service limit.
- check-in bounds long `persona`/`systemPrompt` scalars before aggregate trimming.
- compact trackers retain project entries after the shared item budget is exhausted.

Also fixed list-only authoritative resources: comments retain content/author and default newest-first; activities and notifications retain workflow-required fields. Final deployment used `./default_deploy.sh`; production MCP E2E and health checks passed. Independent reviewer: PASS (478 tests).

ChenNima:
## Re-review — follow-up commits `dc34b95` + `14cebf7`

Re-reviewed against the two fixup commits. All five findings are addressed, and the fixes are clean. Ran the suites locally — **287/287 pass** (`src/mcp/__tests__`, `search.service`, `comment.service`) and `tsc --noEmit` is clean.

**How each was resolved:**

1. **`chorus_search` cap (was #1)** ✅ — now passes `limit: 50` into the service, so the `pageSize: 50` / `maximumRows: 50` metadata is truthful. Bonus: `total` now sums `counts` across entity types instead of `rows.length`, so it reflects the real match count.
2. **`total` semantics (was #2)** ✅ — introduced a `detailSource` discriminator on every inventory entry (`single-get` / `list-authoritative` / `discovery` / `aggregate`), and the docs + tool descriptions now state that bounded discovery tools report `total` as the returned candidate count. Explicit contract instead of an ambiguous field.
3. **`chorus_get_unblocked_tasks` unreachable rows (was #3)** ✅ — promoted to true `page` mode: added `skip`/`take` to the `getUnblockedTasks` service and `page`/`pageSize` to the tool. `chorus_get_project_groups` likewise gained real paging. `available_ideas/tasks` `maximumRows` corrected 100 → 50 to match the service's actual `take: 50`.
4. **`chorus_checkin` hard-fail on long scalars (was #4)** ✅ — `serializeBoundedAggregate` now pre-truncates `agent.persona` / `agent.systemPrompt` to 4096 code points before the byte-cap loop, so a bloated persona can't push check-in into a throw.
5. **`compactTracker` dropping projects (was #5)** ✅ — removed the `if (remaining === 0) break;`, so every project stays in the map; only the per-item budget shrinks the row lists.

**One thing worth a heads-up (not a blocker):** commit `dc34b95` also flipped the **offset-mode** `listComments` ordering from `createdAt asc` → `[createdAt desc, id desc]`, and rewrote the prior code comment that said this path's ordering "MUST stay unchanged." That's the right call for the MCP tool (newest-first is what an agent wants), and I confirmed the shared blast radius is safe: the dashboard comment UI uses **cursor mode** (already newest-first), and I found no frontend caller of the offset REST path (`/api/comments` without `cursor`/`limit`). So the only behavioral change is that the offset REST endpoint now returns newest-first — just flagging it since it's a shared service path that a future REST consumer would inherit.

LGTM to merge.

PR #540, opened 2026-09-04
feat(chorus-pi): support async/detached subagent runs + nicobailon coexistence

expoli:
## Summary

`packages/chorus-pi`'s Chorus session lifecycle assumed the `subagent` tool
runs **blocking** (session opens on `tool_call`, closes on `tool_result`).
The nicobailon `pi-subagents` package is the de-facto subagent implementation
for many users, and it launches **async (detached)** by default: `tool_result`
returns immediately with `details.asyncId`, completion arrives later on the pi
event bus. With this PR the extension's session lifecycle supports both:

- **Async completion**: `tool_result` with `details.asyncId`/`details.runId`
  defers session close to `subagent:async-complete` / `subagent:process-terminal`
  events, with a `session_shutdown` sweep as final guard.
- **No double-injection**: tasks that already carry an injected
  `--- Chorus session` block (main-agent wave templates) are never re-injected.
- **Contract-locked extraction**: the run id comes **only** from
  `details.asyncId`/`details.runId` (verified against pi-subagents source);
  a blocking run's worker output or a generic `id`/`prefix` can never
  misclassify it as async (no session leak).
- **Bundled agents tuned for nicobailon**: `agents/*.md` gain
  `async: false` (reviewer verdict + chorus-worker stay foreground),
  reviewers gain `mcpScript` in `tools` and the nicobailon
  `acceptance: { level: "none" }` marker; `package.json` declares
  `pi.subagents.agents: ["./agents"]` so nicobailon discovers them from the
  installed package (no user-level copy).
- **Coexistence documented**: both implementations register a tool named
  `subagent`; pi rejects a duplicate registration, so the README now documents
  the `settings.packages` package-filter exclude
  (`"extensions": ["!extensions/subagent/**"]`) as the deterministic,
  zero-conflict way to keep nicobailon active while retaining chorus.ts session
  hooks and the packaged agents.

The bundled blocking subagent is **unchanged** (it is the default when no
external implementation is installed); only the async path is added.

## Testing

- `test/all.sh` — offline layers all green (static A: 39 checks; unit B +
  extension-event tests, incl. new deferred-close / process-terminal /
  duplicate-event / marker-suppression / close-failure-retry cases).
- Real-environment verification (pi 0.84.4 + pi-subagents 0.64):
  `subagent({ action: "list" })` shows nicobailon agents, `chorus-worker`
  runs foreground with project context inherited and the Chorus session
  workflow injected.

## Notes

- Two reviewer passes completed; final verdict PASS (one non-blocking P2 note
  on external-contract verification — the nicobailon/pi contract points are
  documented with source line references in `lib/lib.ts`).

### DIFF
--- a/packages/chorus-pi/README.md
+++ b/packages/chorus-pi/README.md
@@ -101,14 +101,74 @@ spawn → run → exit within one tool call, so the extension closes the Chorus
 session at `tool_result`. If you instead use the nicobailon `pi-subagents`
 package's `subagent` tool, top-level launches are **async (detached)** by
 default: `tool_result` returns immediately with `details.asyncId` and the run
-completes later. The extension detects this case (`asyncId`/`runId` in details
-or JSON content) and defers session close to `subagent:async-complete` /
+completes later. The extension detects this case (`asyncId`/`runId` in
+`details`) and defers session close to `subagent:async-complete` /
 `subagent:process-terminal` (with `session_shutdown` sweep as a final guard).
 Tasks that already carry an injected `--- Chorus session` block (e.g. a
 main-agent wave template) are never re-injected.
 
-Note: the bundled subagent and nicobailon `pi-subagents` both register a tool
-named `subagent` — install only one of the two to avoid a tool-name clash.
+### Coexistence with nicobailon `pi-subagents`: load-order rule
+
+The bundled subagent (pi's official reference pattern, at `extensions/subagent/`)
+registers a tool named `subagent`. The nicobailon `pi-subagents` package registers
+a tool with the **same name**. pi's extension loader rejects a duplicate tool
+registration with a conflict error (verified on pi 0.84.4:
+`Tool "subagent" conflicts with ...`), so the two cannot both register.
+
+**Recommended setup (keep nicobailon, zero conflicts)**: exclude the bundled
+subagent extension via a package filter in `settings.packages` — pi's package
+entries accept an object form with per-resource glob patterns:
+
+```json
+"packages": [
+  "npm:pi-subagents",
+  {
+    "source": "git:github.com/Chorus-AIDLC/chorus/packages/chorus-pi",
+    "extensions": ["!extensions/subagent/**"]
+  }
+]
+```
+
+This keeps `chorus.ts` (session hooks) and the `agents/*.md` files (discovered
+via `pi.subagents.agents`) while the bundled `subagent` tool never registers —
+no conflict error, nicobailon wins deterministically.
+
+| Setup | What happens |
+|-------|--------------|
+| Only `@chorus-aidlc/chorus-pi` (no external subagents) | Bundled subagent registers and handles dispatch (single/parallel/chain, blocking) |
+| Both installed, with the filter above | nicobailon's `subagent` tool is the only one. Chorus session hooks keep working (they match on the tool name) |
+| Both installed, no filter: `npm:pi-subagents` listed **before** chorus-pi | nicobailon wins; the bundled subagent reports a conflict error at load (harmless inside an interactive session, noisy for CLI commands like `pi packages list`) |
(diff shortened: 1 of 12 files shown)


expoli:
## Why the packaged agents do not set `async: false` (and what we hit in real use)

The original draft added `async: false` to the bundled agents so that a
reviewer/worker spawn returns the result inline (mirroring the bundled
subagent's blocking semantics). Real-environment testing on **pi-subagents
0.65.0 + pi 0.84.4** showed that is a trap for these agents:

```
Agent 'chorus-task-reviewer' ran as a foreground child, which never loads the
parent's ambient extensions, and these child tools were unavailable: mcp, mcpScript.
...
Agents that need MCP tools ... must run as background children (async: true)
```

### Why

- nicobailon `async: false` produces a **parent-hosted in-process child**
  (`host: "parent"`, shares the parent process). Extensions are process-level
  in pi, so an in-process child cannot load the parent's ambient extensions a
  second time — tools registered by `pi-mcp-adapter` (`mcp`, `mcpScript`) and
  the Chorus MCP servers (`chorus_*`) simply do not exist in the child.
- nicobailon 0.65 checks the agent's `tools` allowlist against the child's
  actual tool set (`evaluateChildToolDiagnostic`) and folds a mismatch into
  the run result at settle — **the run is marked failed (exit 1) even when
  the agent completed its work** (in our case the review read the task,
  inspected diffs, ran tests and posted its VERDICT via `chorus` CLI; the
  child never touched `mcp`).
- The Chorus reviewers/worker need `mcp` (to post verdicts / check in), so a
  foreground run is either a failed run or a silently degraded one
  (bash + CLI workaround).

### What we changed

- `agents/*.md`: no `async: false` — the 4 packaged agents run as nicobailon's
  default **background children** (full ambient extensions, `chorus_*` +
  `mcpScript` available; verified: background child exposes 83 `chorus_*`
  tools and completes cleanly).
- Callers awaiting an inline result use the run notification / `bg_wait`
  instead of foreground blocking.
- README documents the constraint.

### Notes

- This is a nicobailon 0.65 behaviour (0.64 had no such child-tool diagnostic;
  there the same foreground worker silently ran without `mcp`).
- Reported upstream: see the issue draft in
  https://github.com/nicobailon/pi-subagents/issues (capability asymmetry:
  the same agent has a different tool surface foreground vs. background, and a
  completed run is marked failed for a declared-but-missing ambient tool).

expoli:
### Follow-up: implementation comparison (why the bundled subagent can block AND use MCP)

The bundled subagent's file header says it "spawns a separate pi process for
each subagent invocation" — every call is a **full `pi` CLI process**
(`spawn(pi, ["--mode","json","-p","--no-session"])`, index.ts:300/346), which
loads all configured extensions normally and then blocks on process exit.
nicobailon's `async: false` is the opposite: a **parent-hosted in-process
child** (`host: "parent"`) chosen for speed (no process spawn, reuses the
parent's model runtime / MCP connections), which by design cannot load
process-level ambient extensions a second time.

| | Bundled subagent (blocking) | nicobailon `async: false` | nicobailon background (default) |
|---|---|---|---|
| Child form | **spawns a separate `pi` process** | in-process (shares parent process) | separate runner process |
| Ambient extensions | ✅ loaded normally | ❌ not loaded | ✅ loaded |
| MCP tools (`mcp` / `chorus_*`) | ✅ full | ❌ builtin only | ✅ full |
| Blocking semantics | awaits child process exit (natural) | in the event loop (fastest) | run notification / `bg_wait` |
| Cost per call | high (process startup/teardown) | low | medium |

So "blocking + MCP" in the bundled subagent is not an in-process trick — it
takes the same full-process route as nicobailon's background runner. The
trade-off is nicobailon's own: foreground in-process is a performance choice
whose cost is the ambient-tool surface. This PR therefore keeps the 4 packaged
agents on nicobailon's **background** path (full capability, wait via
notification/`bg_wait`), which is the capability-equivalent of the bundled
subagent, and documents the constraint in the README.

expoli:
下面是对应的执行示例
- 主会话面板
image
- subagent 工作面板
image

PR #550, opened 2026-09-10
feat(daemon): per-agent model and thinking level in agents[]

### LINKED ISSUE #549: Daemon: per-agent model and thinking/reasoning effort in `agents[]`
Hi maintainers,

We run a single `chorus daemon` with several agents (`agents[]` in `~/.chorus/daemon.json`) — one per role, each with its own `cho_` key/persona (developer, reviewer, orchestrator). The multi-agent daemon covers *identity* well (key, backend, cwds, permission mode, concurrency), but there is no way to give an individual agent its own **model** or its own **thinking / reasoning effort**: both are whatever the machine-global harness config resolves to. We'd like to propose two optional per-agent fields.

## Current behavior (verified against `main` @ v0.17.3, commit 8cc534fa)

- `resolveAgentConfigs` reads a fixed field whitelist per `agents[]` entry — `url`, `apiKey`, `agentType`, `cwds`, `permissionMode`, `maxConcurrency`, `sigintTimeoutMs`, `browseRoots`, `daemonWake`, `label`, `agentUuid`, `agentName` (`cli/daemon-config.mjs:317-330`, merge at `389-500`). A `"model"` / `"thinking"` / `"env"` key is **silently ignored** — no validation error, so it is easy to believe it took effect.
- The per-agent deps handed to each agent runtime carry `permissionMode` / `agentType` / `cwds` / `maxConcurrency` / `sigintTimeoutMs` / `browseRoots` — no model, no reasoning knob, no env (`cli/daemon.mjs:662-679`); `selectSpawner` forwards only `logger / permissionMode / creds / bundleVersion` (`cli/spawner-select.mjs:61-90`).
- No spawner passes a model **or** an effort/thinking flag:
  - claude: `-p --output-format stream-json --verbose [--session-id|--resume] [--mcp-config] [--dangerously-skip-permissions]` (`cli/claude-spawner.mjs:198-207`)
  - pi: `--mode json --session-id <id> -p` (`cli/pi-spawner.mjs:61-63`)
  - codex: `exec [resume <id>] --json …` (`cli/codex-spawner.mjs:103-110`)
  - kiro: `chat --no-interactive --agent chorus …` (`cli/kiro-spawner.mjs:69-76`)
- The spawned child env is `{ ...process.env, CHORUS_DAEMON_HEADLESS, CHORUS_URL, CHORUS_API_KEY, CHORUS_AGENT_PROFILE }` (`claude-spawner.mjs:356-358`, `pi-spawner.mjs:196-204`, `kiro-spawner.mjs:280-284`) — process-global plus that agent's Chorus credentials, with no per-agent env surface. The only model/effort levers today are therefore per machine: `ANTHROPIC_MODEL` + `CLAUDE_CODE_EFFORT_LEVEL` / `~/.claude/settings.json` `model` + `effortLevel` (claude-code), `$CODEX_HOME/config.toml` `model` + `model_reasoning_effort` (codex), `CHORUS_DSH_MODEL` / `DSH_MODEL` (dsh, `cli/dsh-spawner.mjs:186-190`), pi's own settings (pi).
- `chorus-init` deliberately never persists model-provider material into `daemon.json` (`openspec/specs/chorus-init/spec.md:151`) — correct for *secrets*, but there is currently no path for a non-secret model id or effort level either.
- The one explicit model knob that exists is the foreground launcher passthrough — `chorus agents run --name X -- --model opus` (`docs/DAEMON.md:20-45`, `openspec/specs/cli-agent-launch/spec.md:70-80`) — and it does not apply to daemon wakes.

Net effect: to put one agent on an expensive deep-reasoning configuration and another on a cheap/fast one you must either run separate daemons (different `HOME`, because `daemon.json` / `daemon.pid` hang off `homedir()` — `cli/credentials.mjs:27`, `cli/daemon-lifecycle.mjs:44-51`) or rely on claude-code-only project settings (`.claude/settings.json` inside each agent's served cwd). Both are workarounds, and neither works for codex/dsh/kiro agents.

## Proposal

Add two optional, non-secret strings per `agents[]` entry (`model`, `thinking`), with the top-level values acting as defaults exactly like every other per-agent field, and deliver each through the backend's own mechanism. The daemon stays a **pass-through**: it never normalizes vocabularies, it forwards the value verbatim and lets the harness accept or reject it.

| `agentType` | `model` delivery | `thinking` delivery |
|---|---|---|
| claude-code | `--model <value>` — Claude Code documents `--model` as higher precedence than the settings `model` field (managed/project settings can still restrict via `availableModels`) | `--effort <low\|medium\|high\|xhigh\|max\|ultracode>` — an explicit launch choice sits at step 1 of Claude Code's effort resolution order (above the settings `effortLevel` key and the model default). This matters for us specifically because the daemon wakes are headless `-p` runs, where the docs state a non-interactive `/effort` reports *Not applied* and `--effort` at launch is the correct lever. A managed `maxEffortLevel` cap still applies, and a level the active model doesn't support falls back to the highest supported level at or below it |
| pi | `--model <value>` — `pi --help` → `--model <pattern>` (`provider/id`, optional `:<thinking>`) | `--thinking <off\|minimal\|low\|medium\|high\|xhigh\|max>` — verified in `pi --help` (note `--model id:<level>` is an equivalent shorthand; a `--thinking` flag is cleaner than teaching the daemon pi's model-string grammar) |
| codex | `-m <value>` — verified locally: `codex --help` → `-m, --model <MODEL>` | `-c model_reasoning_effort="<value>"` — the key is verified in a real `~/.codex/config.toml`; the accepted value set comes from the model catalog (`supported_reasoning_levels[].effort`, e.g. `low`/`high`/`max`), so pass through verbatim |
| dsh | `initialize.params.model` — already the field the spawner fills from env today (`cli/dsh-spawner.mjs:409`) | TBD — the `initialize` request the spawner sends today carries only `{ cwd, provider, model }`; needs confirmation that the dsh JSON-RPC contract exposes a reasoning/thinking knob |
| kiro | TBD — `kiro-cli chat` in this repo passes no model flag; needs confirmation that kiro-cli exposes one, otherwise document `model`/`thinking` as unsupported for kiro | TBD (same) |

Design alternatives worth settling in the same pass:

1. **Per-agent `env` map** merged over the inherited environment for **that agent's child only**, with the fail-closed rule already used for credentials — `CHORUS_URL` / `CHORUS_API_KEY` / `CHORUS_AGENT_PROFILE` stay daemon-owned and are applied *after* the merge, so no agent entry can repoint another agent's identity. This covers `ANTHROPIC_MODEL` / `CLAUDE_CODE_EFFORT_LEVEL` / `CHORUS_DSH_MODEL` without new backend-specific code, but leaves codex/pi without a first-class path.
2. **Per-agent extra argv** (a `args: [...]` passthrough appended to the backend's argv), mirroring what `chorus agents run --` already does. One field generalizes model + thinking + future harness knobs, at the cost of an unbounded surface. If maintainers prefer an allowlist over a table that has to track every backend flag, this is probably the smallest maintenance delta.

## Suggested acceptance criteria

- An `agents[]` entry with `"model"` and/or `"thinking"` produces the backend-specific flag for **that agent only**; a sibling agent without those fields keeps today's argv/env exactly.
- Absent / empty values ⇒ byte-for-byte today's behavior; existing `daemon-multi-agent` tests pass unchanged.
- Validation follows the existing resolver contract (`openspec/specs/daemon-multi-agent/spec.md`): non-empty string, otherwise ignored or an error — never a silent fallback to a different model or effort. Unknown values are passed through verbatim and left for the harness to reject, with the failure visible in the daemon log / turn transcript.
- Startup output lists each agent's model and thinking level (the per-agent line already prints backend / permission mode / paths / concurrency — `cli/daemon.mjs:926-929`).
- Spec + docs + tests updated: `openspec/specs/daemon-multi-agent/spec.md` per-agent field list, `docs/DAEMON.md` per-agent field table and per-backend delivery section, plus unit tests for resolver merge/override and per-spawner argv.

## Non-goals

- Choosing or restricting models/effort server-side / in the Chorus UI — the `Agent` model has no such field, and model choice is a client-harness concern.
- Persisting model-provider secrets (`ANTHROPIC_*`, Bedrock, …) into `daemon.json` — remains out of scope per the chorus-init spec.
- Normalizing one cross-backend effort vocabulary (claude `low…max`, pi `off…max`, codex catalog-driven): the values differ per backend and evolve with the harnesses, so the daemon should forward what the operator wrote.
- Changing default models/effort or shipping a model catalog in the daemon.

Happy to open a PR (resolver fields + per-spawner delivery + docs/tests) if the direction looks right — this is a limitation we hit in daily use.

expoli:
Implements the per-agent model + thinking/reasoning-effort selection requested in #549.

One `chorus daemon` process can already serve N independent agents (own key, cwds, backend, permission mode, concurrency) but not their own model or reasoning level: those leaked to machine-global harness config, so a deep-reasoning agent and a cheap/fast one required **two** daemons. Worse, writing the obvious keys into an `agents[]` entry did nothing **and said nothing** — `resolveAgentConfigs` reads a fixed field whitelist, so `"model": "opus"` was silently dropped.

## What this adds

Two optional, non-secret per-agent fields — `model` and `thinking` — with the same top-level-as-default merge semantics as every other per-agent field, resolved on **both** daemon shapes (the `agents[]` path and the flat single-agent path, which resolves through a new `resolveFlatModelFields` because it bypasses `resolveAgentConfigs`) and delivered as each backend's own flags:

| `agentType` | `model` → | `thinking` → |
|---|---|---|
| `claude-code` | `--model <v>` | `--effort <v>` |
| `pi` | `--model <v>` | `--thinking <v>` — inserted **before** pi's trailing `-p`, which must stay last |
| `codex` | `-m <v>` | `-c model_reasoning_effort=<v>` — in both the fresh `exec` and `exec resume <id>` shapes |
| `dsh` / `kiro` / `offline` | not delivered | not delivered |

- **Structural validation only.** A present value must be a non-empty string, else the daemon exits non-zero naming the agent **and** the field (mirroring the per-agent `permissionMode` precedent). The *value* is never checked against a list — the vocabulary is backend-specific and moves with each harness release — so it is forwarded verbatim and the harness accepts or rejects it.
- **No silent drop for unsupported backends.** An agent whose backend cannot receive the fields gets one visible `⚠` startup warning naming the agent, the field, and the backend.
- **Back-compat.** Absent fields leave the spawn argv byte-identical to before; an existing `daemon.json` needs no edits.
- **Foreground parity.** `chorus agents run` applies the same fields with the same mapping; a flag passed after `--` wins (both the separated and `=` spellings are detected).
- **Observability.** The multi-agent startup line appends `model=` / `thinking=`; the single-agent banner gains `Model` / `Thinking` rows.

## Evidence

- `cli/__tests__/daemon-per-agent-model-thinking.test.mjs` (new) plus extended `daemon-multi-agent-config`, `{claude,pi,codex}-spawner`, `spawner-select`, `agent-launcher` and `daemon-multi-agent-runtime` suites — including a cross-surface test that slices the real spawner argv builders and compares them to the launcher's mapping, so the two surfaces cannot drift.
- Full CLI suite: **89 files / 1580 tests green**.
- Real `chorus daemon` run with six configured agents: each printed its own resolved values, the kiro agent got exactly one warning, a field-less agent spawned with **no** model flag, and two `pi` agents completed live turns (`exitCode 0`) with their own `--model`/`--thinking` in argv.
- The server-side chain too: a human @-mention in the UI → SSE `mentioned` → directed to the right connection (siblings logged “not this daemon … suppressing broadcast wake”) → wake → spawn → `wake done (exit=0, 61.6s)` → turn ended.

## Known limitations

- `dsh` / `kiro` are not supported in v1 (no parameter could be verified for either) — they warn rather than silently ignore. `dsh`'s `initialize.params.model` is the obvious extension point.
- `chorus agents run` selects from an `agents[]` entry, so a **flat** `daemon.json`'s top-level `model`/`thinking` apply to daemon wakes only (documented).

## Docs / spec

`docs/DAEMON.md` gains the two field rows, the merge/validation contract, a per-backend delivery section and the unsupported-backend caveat. Spec deltas for `daemon-multi-agent` and `cli-agent-launch` are included and archived (`openspec/changes/archive/2026-09-10-add-daemon-per-agent-model-thinking/`).

Closes #549.


ChenNima:
I'll add a unified ENV/params config for wilder usage

expoli:
Understood on the unified ENV/params direction — thanks for saying so plainly rather than leaving it hanging. We keep the explicit fields in our local fork (in daily use here), so there is no urgency from our side; happy to rebase onto whatever shape you land, or to take the params half if you want a second pair of hands.

Because we did the per-backend verification while working #549, here are the facts that bear on an ENV-vs-params split, in case they save you a pass:

- **claude-code** — a model env knob does exist (`ANTHROPIC_MODEL`), but for a *headless* run the launch flag is what actually applies: an explicit launch choice such as `--effort` outranks the settings `effortLevel` and the model default, whereas a non-interactive `/effort` reports "Not applied" in `-p` runs (project/managed settings still re-apply each launch). Also: gateway deployments can reject the ids the `_DEFAULT_*_MODEL` aliases point at (`403 … key not allowed to access model`); passing an explicit allowed id via `--model` is what unblocked us.
- **pi** — no env-var model selector we could verify (`PI_MODEL` is an *output* variable). Model + level are launch flags: `--model`, `--thinking`, and `-p` must stay the last argument.
- **codex** — the model comes from `$CODEX_HOME/config.toml`; the per-invocation overrides are `-m` and `-c model_reasoning_effort=<level>`. So a per-agent ENV map alone would not cover codex/pi model selection unless each agent also got its own `CODEX_HOME`.

Machines-readable evidence from the closed PR (argv per backend, live turns, the directed-wake chain) is in the #549 thread if it is useful for the implementation.

PR #551, opened 2026-09-10
test(cli): isolate the CLI suite from the real ~/.chorus state

expoli:
Two ways the CLI suite depended on the *developer's* real `~/.chorus` state. Both are invisible in CI (where that directory does not exist), so they only surface once someone actually runs the daemon — i.e. the more the project is used, the more likely they are to bite.

## 1. The suite wrote the developer's real usage map

`codex-backend-integration.test.mjs` injected a temp path for the codex *session* map but let the *usage* map fall back to its default, so running the suite created (or overwrote) `~/.chorus/codex-usage.json` with fixture data — observed key `33333333-…`, thread `019f091a-…`. Three spawner constructions were affected.

Fix: every usage-map call in that file is pointed at an isolated temp file, so the real capture/normalize code path stays exercised. In addition the file now snapshots the **real** path's content in `beforeAll` and asserts it is unchanged in `afterAll` — comparing content rather than asserting absence, so a developer who legitimately has that file cannot get a false positive.

The guard was verified by removing one injection: the real file reappeared and the guard failed (`expected '{…33333333…' to be null`); restoring it turned the file green again.

## 2. Five `runDaemon` suites read the developer's real `daemon.json`

`daemon-claude-notfound-warning`, `daemon-credential-completion`, `daemon-integration`, `daemon-lifecycle-dispatch` and `daemon-permission-wiring` build their deps inline with no `loginPath` / `readJson` override, so `runDaemon` read the machine's real `~/.chorus/daemon.json`. Once that file contains an `agents[]` array, `hasConfiguredAgents()` sends them down the multi-agent branch and 14 assertions fail.

Fix: each of the five suites points `HOME` at its own temp dir (`beforeAll`/`afterAll`), which isolates every HOME-derived reader — `daemon.json`, the codex maps, `~/.claude/settings.json` — rather than patching one reader per test.

Measured on a pristine `origin/main` checkout with a real two-agent `daemon.json` present:

| | files | tests |
|---|---|---|
| before | 5 failed / 83 passed | 14 failed / 1504 passed |
| after | 89 passed | 1580 passed |

(Unrelated pre-existing failures in four `src/**` files in this environment — jsdom `localStorage` unavailable and a zh locale — reproduce identically on a pristine `origin/main` checkout and are not touched by this PR.)

### DIFF
--- a/cli/__tests__/codex-backend-integration.test.mjs
+++ b/cli/__tests__/codex-backend-integration.test.mjs
@@ -11,15 +11,20 @@
 //      CodexSpawner; driving its wake() through a fake spawn yields the expected
 //      `codex exec --json` (new) and `codex exec resume <id>` (known anchor) argv,
 //      prompt on stdin, daemon URL/key in the child env (never argv).
-import { describe, it, expect, vi } from "vitest";
+import { describe, it, expect, vi, beforeAll, afterAll } from "vitest";
 import { EventEmitter } from "node:events";
-import { mkdtempSync, rmSync } from "node:fs";
+import { existsSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { buildDaemon } from "../daemon.mjs";
 import { ClaudeSpawner } from "../claude-spawner.mjs";
 import { CodexSpawner } from "../codex-spawner.mjs";
 import { getThreadId, setThreadId } from "../codex-session-map.mjs";
+import {
+  codexUsageMapPath,
+  getCodexUsageSnapshot,
+  setCodexUsageSnapshot,
+} from "../codex-usage-map.mjs";
 import { Waker } from "../waker.mjs";
 import { createTranscriptUploadHooks } from "../upload-hooks.mjs";
 import { killProcessTree } from "../process-killer.mjs";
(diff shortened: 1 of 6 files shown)


ChenNima:
Reviewed — merging. The problem class is well chosen: state leakage into the developer's real `~/.chorus` is invisible in CI (that directory does not exist there), so it only ever surfaces for people who actually run the daemon. The negative test (removing one injection, watching the real file reappear and the guard fail, then restoring it) is exactly the evidence standard we want for a guard, and comparing content rather than asserting absence avoids false positives for developers who legitimately have the file. I also checked the three remaining `runDaemon` suites — `chorus-entry-daemon-d`, `daemon-multi-agent-runtime`, `daemon-install-integration` — and they were already isolated via injection or a throwaway HOME, so the coverage here is complete. The default pool is `forks`, so the `process.env.HOME` mutation cannot leak across files.

Retargeting to `develop`: this repo's flow is feature → develop → main, `main` only receives release merges.

Four follow-up notes, none blocking:

1. **`codex-backend-integration.test.mjs` could use the same temp-HOME trick as the other five.** `codexUsageMapPath()` resolves through `homedir()`, so pointing HOME at a temp dir would isolate the usage map in ~3 lines instead of ~30 lines of per-call `getUsageSnapshotFn` / `setUsageSnapshotFn` injection, still exercise the real capture/normalize path, and make every *future* test in that file immune without anyone remembering to inject. As written, the easy-to-miss step stays where it was.

2. **Five copy-pasted temp-HOME blocks want a shared helper** (e.g. a `useTempHome()` in a `cli/__tests__/__utils__/`). Because CI has no `~/.chorus`, a regression of this class stays silent — a helper is what makes the next contributor get isolation by default rather than by memory.

3. **The temp HOME does not isolate on Windows.** `os.homedir()` reads `USERPROFILE` there, not `HOME`. Our CI is ubuntu-only so the gate is unaffected, but the npm package targets Windows, and a Windows developer running the suite gets none of this fix. Setting `USERPROFILE` alongside `HOME` covers it.

4. **The content guard asserts on a machine-global path**, so it can fail for the wrong reason — a live local daemon, or another concurrently forked suite, writing `~/.chorus/codex-usage.json` during the run would make this file report the failure. Folding it into note 1 removes the hazard entirely.

Thanks for the careful before/after measurement on a pristine checkout, and for separating out the pre-existing unrelated `src/**` failures.

PR #552, opened 2026-09-10
docs(openspec): align the specs of record with pi being wakeable

expoli:
`add-daemon-pi-backend` made `pi` a first-class wakeable backend (`agent-type-map.mjs` maps `pi` → `"pi"`, `KNOWN_AGENTS` includes it, `selectSpawner` has an explicit `PiSpawner` branch, `isWakeableAgentType("pi")` is true) and updated the `pi-daemon-backend` capability — but left four older statements describing `pi` as non-wakeable/`offline`. The spec of record therefore contradicted both itself and the code, and a reader (or a future change) following the stale text would classify a pi agent as never-woken.

Corrected (spec-only — no runtime change):

- `daemon-multi-agent` / **Per-agent backend selection** — the example backend list gains `pi`, plus a scenario pinning the claude-code + pi pairing.
- `daemon-multi-agent` / **Offline agent type is a valid, never-woken agents[] entry** — the fail-closed list drops `pi` (it stays for `opencode`, `openclaw`, and `dsh` while its backend is de-advertised) and says so explicitly.
- `chorus-init` / **Daemon backend agentType derived from the init selection** — states `pi` → `pi`, drops `pi` from the fail-closed list, stops using `pi` as the “non-wakeable” example, and gains a scenario pinning the pi mapping.
- `chorus-init` / **Per-agent daemon-wake defaults off at init** — the wakeable-backend list gains `pi`; the offline scenario no longer lists it.
- `docs/DAEMON.md` — the per-agent `agentType` row lists the full accepted set, and the intro no longer claims the daemon wakes only a headless Claude Code.

`MODIFIED` blocks carry the full requirement text plus every retained scenario, as OpenSpec requires.

## Verification

- Exhaustive grep for each stale phrasing across `openspec/specs` and `docs`: `opencode, openclaw, pi` → 0, `pi, and dsh` → 0, `opencode / openclaw / pi` → 0, `opencode`…`or`…`pi` → 0.
- `openspec validate fix-pi-wakeable-spec-drift` passed; the change is archived as `2026-09-10-fix-pi-wakeable-spec-drift` and merged into the cumulative specs (+2 requirements, ~2 modified).
- The corrected statements are already pinned by existing tests for the code they describe (`init-agent-type-map`, `daemon-agent`, `spawner-select`, `pi-spawner`), which assert `pi` → `pi` and `isWakeableAgentType("pi") === true`.

### DIFF
--- a/cli/init/steps/credential-seed.mjs
+++ b/cli/init/steps/credential-seed.mjs
@@ -68,7 +68,7 @@ function nonEmpty(value) {
  * Whether an init selection id is the DeepSeek Harness (dsh). dsh needs a SECOND
  * credential sink beyond ~/.chorus/daemon.json — see {@link writeDshCredentialsEnv}.
  * We gate on the selection id, NOT the daemon agentType: dsh maps to the shared
- * "offline" agentType (agent-type-map.mjs) alongside opencode/openclaw/pi, so the
+ * "offline" agentType (agent-type-map.mjs) alongside opencode/openclaw, so the
  * agentType cannot single dsh out. The id can.
  * @param {string} id
  */
--- a/docs/DAEMON.md
+++ b/docs/DAEMON.md
@@ -2,8 +2,9 @@
 
 The Chorus daemon is a local, long-lived client. It connects to a remote Chorus
 server, subscribes to the agent notification stream, and wakes a local headless
-Claude Code on task dispatch — so an assigned agent can act on work even when no
-one is at a terminal.
+coding agent on task dispatch — so an assigned agent can act on work even when no
+one is at a terminal. The backend is selectable per agent and defaults to
+Claude Code.
 
 ```bash
 npx @chorus-aidlc/chorus daemon          # foreground
@@ -194,7 +195,7 @@ agent **overrides** it for that agent only. Per-agent fields:
 |-------|---------|
 | `apiKey` | *(required)* the agent's `cho_` key — determines its identity |
 | `url` | Chorus server (may differ per agent — different server/company) |
-| `agentType` | `claude-code` \| `codex` \| `kiro` (backends may be mixed) |
+| `agentType` | `claude-code` \| `codex` \| `kiro` \| `pi` \| `dsh` \| `offline` (backends may be mixed; `offline` is never woken) |
 | `cwds` | working directories this agent serves (one connection each) |
 | `permissionMode` | `yolo` \| `chorus` |
 | `maxConcurrency` | this agent's own wake-concurrency cap (default `4`) |
--- a/openspec/changes/archive/2026-09-10-fix-pi-wakeable-spec-drift/.openspec.yaml
+++ b/openspec/changes/archive/2026-09-10-fix-pi-wakeable-spec-drift/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-10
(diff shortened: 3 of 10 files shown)


ChenNima (changes requested):
Verified every claim in this PR against the code and they all hold: `cli/init/agent-type-map.mjs:34` maps `pi` → `pi` (and `dsh` → `offline`), `isWakeableAgentType` excludes only `offline`, `cli/spawner-select.mjs:83` has the explicit `PiSpawner` branch, and `cli/agent-backend-prompt.mjs:44` advertises `pi` in the menu. The `docs/DAEMON.md` row listing `dsh` is correct too, not a slip — `KNOWN_AGENTS` includes it and it stays reachable by name while the backend is de-advertised.

What I like: this fixes a real self-contradiction in the specs of record rather than polishing wording; the `MODIFIED` blocks carry full requirement text plus every retained scenario; the change was archived before the PR (matching our archive-before-PR convention); and instead of only deleting stale words you added two scenarios that pin the behavior (`pi` selection → `pi` agentType, and claude-code + pi coexisting in one daemon), with the note that existing tests already assert them. The reverse pointer in the `offline` requirement ("`pi` is NOT among them … see the `pi-daemon-backend` capability") is a nice touch — the next person tempted to put `pi` back in the fail-closed list hits that sentence first.

Requesting changes for one leftover of exactly the drift this PR claims to have eliminated:

**`cli/init/steps/credential-seed.mjs:71`**

```
 * "offline" agentType (agent-type-map.mjs) alongside opencode/openclaw/pi, so the
```

Still describes `pi` as mapping to `offline`. It is a load-bearing comment — it explains why `isDshSelection` gates on the selection id rather than the agentType — and the argument survives, only the `pi` member is wrong. Dropping `pi` from that list is the whole fix.

The "exhaustive grep" in the description has two blind spots worth noting so this does not recur: the scope was `openspec/specs` + `docs` only, so `cli/**` comments were never scanned; and the patterns (`opencode, openclaw, pi`, `pi, and dsh`, `opencode / openclaw / pi`) all assume spaced separators, so the compact `opencode/openclaw/pi` form slips through. A pattern like `-nEi "openclaw[,/ ]+pi"` over `cli docs openspec/specs public` catches both classes.

The stale phrasings under `openspec/changes/archive/**` are correctly left alone — archived changes are historical records and should keep the text as it was.

Also retargeting this to `develop`: our flow is feature → develop → main, and `main` only takes release merges.

expoli:
Done — thanks for the precise catch and for the method feedback; both are applied.

**The fix** (`8883ac07`): dropped `pi` from the `offline` list in `cli/init/steps/credential-seed.mjs`. The comment's argument is untouched — gating on the selection id because the `agentType` cannot single out `dsh` still holds, since `opencode` / `openclaw` also map to `offline`; only the `pi` member was wrong.

**Both scan blind spots were real**, and worth stating plainly: the scope was specs + docs, so `cli/**` comments were never read (code comments are documentation and go stale the same way), and my patterns assumed spaced separators. Re-scanned every root with your style of pattern (`-e 'openclaw[,/ ]+pi'`) plus a broad `pi` + offline/non-wakeable sweep over `cli src public packages plugins docs openspec/specs .agents`, archived changes excluded. That was the only remaining stale instance — the other hits were the corrected spec text, the wakeable-`pi` assertions in `init-integration.test.mjs` / `init-credential-seed.test.mjs`, and false positives (`pinned`, `compiled`).

**Base**: retargeted to `develop`.

Also: thanks for verifying that the `docs/DAEMON.md` `dsh` row is correct rather than filing it as noise — that saved a round trip.

ChenNima (approved):

Issue #555, opened 2026-09-11
Concurrent pm_update_* draft updates silently lose data (read-modify-write on the whole JSON column)


expoli:
## Summary

`chorus_pm_update_document_draft` / `chorus_pm_update_task_draft` return success, but concurrent updates to sibling drafts of the same proposal can be **silently lost**. The service layer implements each call as a read-modify-write of the **entire `documentDrafts` / `taskDrafts` JSON column**, with no locking or version check, so two in-flight requests read the same snapshot and the last write wins.

## Environment

- Chorus server v0.17.3 (local, PGlite); source `main` @ `d30af3a1`
- Client: `@chorus-aidlc/chorus-pi` 0.17.3 (MCP tool surface; same backend via `chorus mcp call`)

## Evidence (source)

`src/services/proposal.service.ts`:

- `updateTaskDraft` (line 1242): reads `proposal.taskDrafts` (1256), mutates one element in memory (1274), writes the **whole array** back (1277-1280, `taskDrafts: existingDrafts`).
- `updateDocumentDraft` (line 1204): identical shape for `documentDrafts` (read 1218 / mutate 1225 / write 1230).
- `addDocumentDraft` (1144), `addTaskDraft` (1182), `removeDocumentDraft` (1304), `removeTaskDraft` (1337) share the same whole-array write pattern.

```ts
// src/services/proposal.service.ts:1242 (updateTaskDraft, abridged)
const proposal = await prisma.proposal.findFirst({ where: { uuid: proposalUuid, companyUuid, status: "draft" } });
const existingDrafts = (proposal.taskDrafts as unknown as TaskDraft[]) || [];
existingDrafts[draftIndex] = { ...existingDrafts[draftIndex], ...appliedUpdates };
await prisma.proposal.update({
  where: { uuid: proposalUuid },
  data: { taskDrafts: existingDrafts as unknown as Prisma.InputJsonValue },
});
```

Two concurrent calls both read snapshot N; each writes back its own copy of the array; the second commit overwrites the first draft's change. The MCP response only reports `task_draft_updated` / `document_draft_updated` - there is no version/conflict signal, so the loss is invisible without a read-back.

## Reproduction

1. Proposal in `draft` status with >=2 task drafts (or >=2 document drafts).
2. Issue several `chorus_pm_update_task_draft` calls for **different drafts of the same proposal, in parallel** (e.g. separate MCP tool calls in one batch).
[...]

PR #556, opened 2026-09-11
fix(proposal): serialize draft updates under a proposal row lock

L4XB:
Fixes #555.

`documentDrafts` / `taskDrafts` are whole JSON columns, and every draft mutator in `src/services/proposal.service.ts` (`addDocumentDraft`, `addTaskDraft`, `updateDocumentDraft`, `updateTaskDraft`, `removeDocumentDraft`, `removeTaskDraft`) was a read-modify-write of the entire list with no lock or version check. Two in-flight `chorus_pm_update_*` calls on sibling drafts read the same snapshot, each wrote "its" array back, and the last write silently won while both reported success.

**Change**

- A small `withLockedDraftProposal(proposalUuid, companyUuid, mutate)` helper runs the mutation inside `prisma.$transaction`, takes `SELECT "uuid" FROM "Proposal" WHERE "uuid" = $1 FOR UPDATE` first, then does the `findFirst` (same `companyUuid` / `status: "draft"` scoping as before) and hands the fresh row to the mutator, which writes through the transaction client. Concurrent callers serialize on the row for the length of the transaction, so the second one reads the first one's result. Works the same on Postgres and PGlite (which already runs with a single connection).
- The six mutators keep their validation (acceptance-criteria rules, "draft not found"), their error messages, and emit the change event after the transaction as before. The diff looks larger than it is because each body moved into the callback.

**Tests**

- `src/services/__tests__/proposal-draft-concurrency.test.ts`: a state-backed prisma mock whose `$transaction` serializes callbacks like a row lock. Two concurrent sibling document-draft updates both land, a concurrent task update plus document add both land, the lock is taken inside the transaction before the read, and the not-in-draft error is unchanged. The first two cases fail on `develop`.
- `proposal.service.test.ts`: the prisma mock now implements `$transaction` by calling back with the same mock (and stubs `$queryRaw`), so the existing draft tests keep observing the reads/writes. 124 tests pass across both files.

`eslint` reports only the two pre-existing unused-import warnings in `proposal.service.ts`.

### DIFF
--- a/src/__tests__/integration/acceptance-criteria-enforcement.integration.test.ts
+++ b/src/__tests__/integration/acceptance-criteria-enforcement.integration.test.ts
@@ -35,6 +35,8 @@ const { mockPrisma, mockEventBus, mockFormatCreatedBy, mockFormatReview, mockIsA
       createMany: vi.fn(),
       deleteMany: vi.fn(),
     },
+    $queryRaw: vi.fn(async () => []),
+    $transaction: vi.fn(async (fn: (tx: unknown) => Promise<unknown>) => fn(mockPrisma)),
   };
   return {
     mockPrisma,
--- a/src/services/__tests__/proposal-draft-concurrency.test.ts
+++ b/src/services/__tests__/proposal-draft-concurrency.test.ts
@@ -1,0 +1,129 @@
+/**
+ * Concurrent draft updates on one proposal must not lose each other (#555).
+ *
+ * `documentDrafts` / `taskDrafts` are whole JSON columns. Each mutator reads
+ * the list, changes one element and writes the whole list back; two in-flight
+ * calls that read the same snapshot let the last write win. The mutators now
+ * run inside a transaction that locks the proposal row first, so the second
+ * caller reads the first caller's result.
+ */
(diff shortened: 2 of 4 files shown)


ChenNima:
Maintainer follow-up pushed in f2acbb22.

Changes:
- derive the transaction callback type from the actual `$extends` Prisma client (fixes TS2345);
- make the concurrency tests exercise add/update and update/remove on the same JSON column;
- add the transaction/raw-query stubs missed by the acceptance-criteria integration test.

Local verification:
- `pnpm exec tsc --noEmit`
- 132 targeted tests
- full coverage suite: 6,381 passed, 16 skipped
- release-contract and plugin shell suites passed

Adjacent observations, intentionally left out of this focused #555 patch:
- `submitProposal` validates before its status update without holding the same row lock, so a draft mutation can land between validation and transition;
- concurrent `approveProposal` calls can both pass the outer pending-status check and materialize duplicate tasks/documents;
- legacy `updateProposalContent` can still replace whole draft arrays, though it currently has no production callers.

Those lifecycle concerns deserve follow-up hardening rather than expanding this PR without dedicated tests.

PR #569, opened 2026-09-17
fix: converge phantom running turns (retry terminal report, turn-driven interrupt, exit-authoritative spawners)

ChenNima:
Resolves idea [47ae1d67](https://chorus-dev.chorus-ai.dev) — 「Daemon 子进程已退出但 UI 仍显示「运行中」，且无法打断（幽灵 running turn）」.

## The bug

A daemon-woken subprocess ends, but the UI keeps showing the conversation as running and offers no way to stop it. The turn stays `running` forever.

Root cause was **two unrelated things** (investigated from the reporting host's logs and pi's own session jsonl):

1. **The reported pi case never actually exited.** Its session file ends on an unanswered `bash` toolCall (`psql … | head`) with `timeout` omitted — pi's bash tool has **no default timeout** (`dist/core/tools/bash.js`: `"Timeout in seconds (optional, no default timeout)"`). It blocked for 11.4 hours until the daemon's shutdown SIGTERM. Recorded as background only; **no timeout or watchdog is introduced here** (explicit owner decision: 「先不管超时吧，不安全，我无法判断一个turn到底要多久结束」).
2. **The turn could never converge.** The terminal turn-advance was fire-and-forget with zero retry, and `reconcileOrphanTurns` only collects turns whose origin connection is stale >90s — a live, heart-beating daemon is never orphan-eligible.

## What changes

- **`cli/daemon-rest-client.mjs`** — bounded retry on the **terminal** turn-advance edge only: 3 attempts, 500ms then 2000ms; retryable = network failure / 429 / 5xx; every 4xx is a server verdict and returns immediately. `running` and all other operations stay single-shot. Idempotent by construction via `advanceTurnForWake`'s correlated-terminal short-circuit.
- **`cli/control-handler.mjs`** — an interrupt with no live child no longer returns silently; it reports the turn as `interrupted(user)` through the waker's existing turn reporter (same instance, no second transport). Only for `idea` / `daemon_session`, where `sessionId === entityUuid`.
- **`cli/child-exit.mjs` (new) + all five spawners** — settle a wake on process `exit`, not only on stdio `close`. A backgrounded grandchild that inherits the pipes kept finished wakes "running". `close` first still settles immediately; `exit` first waits a bounded 2s for pipe drain, then settles anyway. That grace starts only **after** the process is gone and is explicitly **not** a wake timeout.
- **UI + `POST /api/daemon/control`** — the interrupt control is driven by the **turn**, not by the presence of an execution row, so a phantom turn is always clearable by hand. The server also settles the turn itself when its own state shows no live run can act: connection not live **or** no `running` execution row for the entity.

## Design note worth reviewing

The second half of that gate (`!hasRunningExecution`) covers a **zombie SSE** — reverse channel silently dead while the REST heartbeat keeps `lastSeenAt` fresh, so the connection reads online and the control event goes into a channel nobody is listening on. It uses the execution snapshot the daemon uploads on *every* lifecycle transition as server-owned evidence.

Elaboration round 1 chose `daemon-only` and explicitly rejected `daemon-plus-server`. This settle is narrower than that rejected option: it is **only ever triggered by a human clicking interrupt**, never a background reconcile. A stale-evidence false positive still lands on `interrupted` — the state the human asked for — absorbed by the idempotent terminal short-circuit. Flagging it so the owner can re-confirm.

Also included: a stale terminal execution row alongside a still-`running` turn counts as "no live run", so a row's Resume alone can no longer strand the turn; and the success toast distinguishes "cleared here" from "forwarded to a live daemon" via the new `settled` flag rather than claiming success unconditionally.

## Verification

Unit/component: `env -u CHORUS_AGENT_PROFILE pnpm test` → **6600 passed / 0 failed**; `npx tsc --noEmit` clean; `pnpm lint` byte-identical to the pre-change baseline. (This shell injects `CHORUS_AGENT_PROFILE`, which trips three pre-existing spawner env assertions — unrelated to this change.)

Live end-to-end against the deployed stack (chorus-dev), using an isolated throwaway daemon so the real one was never touched:

| Scenario | Result |
|---|---|
| online + running execution row | `settled:false`, turn stays running — daemon keeps authority |
| online + no running execution row (zombie SSE) | `settled:true`, turn → `interrupted/user` |
| connection offline | `settled:true`, turn → `interrupted/user` |
| terminal turn-advance, first 2 hops fail | 3 attempts / 2588ms / `ok:true`, turn → `ended` |

Real browser on live (Cognito login): a genuine phantom turn shows the control (`Clear this stuck turn`), the confirm dialog uses honest copy ("Nothing is killed — there is nothing left to kill."), and clicking it converged the turn to `interrupted/user`. Screenshots captured in light and dark.

Real subprocess: `bash -c "sleep 300 & echo tail; exit 5"` → `close` never fires, settles on exit at 2003ms with the trailing stdout still parsed and the diagnostic warn emitted.

Daemon-side behaviour diff, old vs new generation, same probe:
```
old: control: no running subprocess … ; ignoring interrupt
new: control: no running subprocess … ; reporting the turn as interrupted(user)
```

## Reviews

Four independent task reviews plus a ship-time code-review gateway, all PASS / PASS WITH NOTES. Two round-1 proposal blockers were real and changed the design: the zombie-SSE gap above, and a missing integration checkpoint. Reviewer notes that were fixed rather than waved through: a redundant `(attempt 1/3)` log line on a non-retried 4xx, a shadowed local, a stale `baseDelayMs` in the tech design, an aria-label that still said "Interrupt this running execution", and the unconditionally optimistic toast.

## Known gaps, deliberately not fixed here

- **Legacy `::` residual sessions**: the route's `sessionId = entityUuid` settle can resolve the *modern* session whose `sessionId === ideaUuid` and clear the wrong turn. Narrow, but a clean fix means touching `advanceTurnForWake`, a chokepoint the daemon path also uses — worth its own change.
- The execution-snapshot upload itself has no retry, leaving a residual double-failure window.
- `docs/design.pen` is **not** updated — waived by the owner for this change (headless has no Pencil MCP access).

OpenSpec change archived as `2026-09-17-fix-phantom-running-turn`; `openspec validate --all` → 139 passed / 0 failed; all three updated cumulative specs mirrored back to their Chorus Documents and verified byte-exact.


ChenNima (commented):
## Review (Pi) — PR #569

Read the full diff (cli + server + UI), re-ran the four touched cli/route suites locally (98/98), CI green. The four changes do what the description says, each is narrowly scoped, and the idempotency argument via `advanceTurnForWake`'s correlated-terminal short-circuit holds. The confirm-dialog / toast copy is honest (`settled` distinguishes "cleared here" from "forwarded"). i18n parity is complete across en/zh/ja/ko.

### One real gap worth a follow-up (non-blocking, narrow)

Both the server settle and the daemon no-child report key on the **exact** `entityType:entityUuid`, but a session's `running` turn can belong to a **sibling wake** anchored on the same idea (`task:T` / `proposal:P` / `document:D` whose `directIdeaUuid === A`; `cli/waker.mjs:439` `sessionId = directIdeaUuid`). The UI's stuck-turn control sends `idea:A` (`sessionControlTarget`).

When the UI has no `running` row for the session (stale UI, or a lost execution snapshot — the acknowledged no-retry gap) while a `task:T` child is actually alive:

- **Server** (`route.ts`): `hasRunningExecution(idea, A)` → `false` (it does not consider `directIdeaUuid`), so the turn is settled `interrupted(user)` while the child is still running.
- **Daemon** (`control-handler.mjs`): `executions.get("idea:A")` → miss → reports `interrupted(user)` for session A **without a turnUuid** → FIFO picks the sibling's running turn. No check that another execution in the same session is alive, and nothing gets killed.

Net: UI says *Interrupted*, the agent keeps working, and the eventual `ended` report lands as `invalid_transition`. So the PR's "stale evidence is safe because the online daemon will kill the child anyway" argument holds only when the control key equals the executing key.

Suggested fix (small, both sides mirror the UI's `executionMatchesSession` rule):
1. `hasRunningExecution`: for `entityType === "idea"` use `OR: [{ entityType: "idea", entityUuid }, { directIdeaUuid: entityUuid }]`.
2. `control-handler` no-child branch: before reporting, scan `waker.executions` for a `running` entry with `directIdeaUuid === entityUuid`; if found, **kill that one** (honours the human's intent) instead of reporting a miss.

Reachable only via a missing/stale row, so I'd take it as a follow-up idea alongside the legacy `::` one rather than block this PR.

### Minor notes
- `control-handler`: a `queued` entry (no child yet) also takes the report path. Harmless when the session has no running turn (`not_found`), but it is the same sibling hazard as above.
- `child-exit.mjs`: after settling on `exit`, late `data` chunks can still reach the ndjson parsers after the wake has been reported terminal. Benign today (transcript hooks are warn-not-throw), just noting the ordering.
- Design-note re: elaboration round 1 (`daemon-only` vs `daemon-plus-server`): I agree the human-click-only gate is materially narrower than the rejected background reconcile; owner sign-off on that boundary is the right call.

**Verdict: LGTM with the follow-up above.** Not merging — owner's call.

ChenNima (commented):
## Re-review (Pi) — commit `3689006e`

Re-read the third commit against my previous findings; re-ran the four affected suites locally (321/321), CI green.

**Sibling-wake gap — fixed correctly on both sides.**
- `hasRunningExecution`: `idea` → `OR [{entityType,entityUuid}, {directIdeaUuid}]`; `daemon_session` unchanged. Mirrors `executionMatchesSession` as suggested.
- `control-handler`: falls back to a `running`+`child` entry with `directIdeaUuid === entityUuid`, kills **that** child, and marks `markInterrupting(killEntityType, killEntityUuid)` on the sibling — the crash-vs-user attribution detail is the part that was easy to get wrong, and it's right. `queued` siblings correctly fall through to the report path.

**Legacy `::` resolver (`resolveControlSessionId`)** — read-only, evidence-based, refuses when ambiguous. I agree with the choice: "settle nothing and log" beats any priority heuristic here, since the wrong guess writes a terminal state onto an unrelated live conversation. Single-candidate short-circuit without a running-turn check is fine (`advanceTurnForWake` answers `not_found`).

**Gate inside `try`** — correct; the early `return success({dispatched:true, settled:false})` inside the try is fine.

**Your reasons for not changing the `queued` and late-stdout notes — accepted.**

### One residual note (non-blocking, very narrow)
The **daemon-side** no-child report (`advanceTurn({sessionId: entityUuid})`, no `turnUuid`) has no equivalent of the server's resolver/ambiguity guard: it always addresses `sessionId === entityUuid` and settles via status-FIFO. Path to a wrong write: legacy conversation `A::conn-1` (conn-1 alive) with a stuck turn, while the **modern** session `A` is genuinely running on conn-2 (same agent, other cwd) *and* its execution row is not visible to the UI. Server side refuses (ambiguous); daemon 1 gets the event, has no child, reports `interrupted(user)` for `A` → FIFO lands on conn-2's live turn. Reachable only with a legacy residual + a missing row, and the UI's all-slice fallback normally shows conn-2's row (so the stuck control wouldn't render). Cheapest guard if you want one later: on the un-correlated terminal FIFO in `advanceTurnForWake`, restrict to turns whose session `originConnectionUuid === connectionUuid` (new daemons always send `turnUuid` for their own wakes, so only this report + legacy daemons use the un-correlated path). Fine as a follow-up alongside the snapshot-upload-retry gap.

**Verdict: LGTM. Ready to merge from my side** — owner's call.

PR #572, opened 2026-09-20
fix(chorus-pi): pin Chorus agents to the background path, and stop asking for blocking

expoli:
## Summary

#540 taught the extension's session lifecycle to handle nicobailon `pi-subagents`' async runs and removed `async: false` from the packaged agents, documenting in the README that the reviewers/worker "run as background children (nicobailon default)". The **instruction layer was never updated with it**: the injected Quick Reference, the three reviewer nudges, three skills and all four agent descriptions still told the agent to use "the blocking `subagent` tool".

So an agent that follows the shipped instructions puts a Chorus agent in exactly the configuration #540 removed — and loses the run:

| dispatch | child | declared `tools: …, mcp, mcpScript` |
|---|---|---|
| background (default) | detached runner process | all tools present — reviewer works |
| `async: false` | in-process `AgentSession` | no `mcp`/`mcpScript` → *these child tools were unavailable* → **run settled failed** |

For the worker (no `tools:` allowlist) nothing fails — it silently loses `chorus_*` and cannot check in, report, or submit, while the run reports success.

This PR does the two halves: **pin the mode** and **stop asking for the wrong one**.

## 1. Mechanism — keep Chorus agents on the background path

`extensions/chorus.ts` now pins a dispatch at `tool_call` whenever a `chorus-*-reviewer` or a worker (`worker`, `chorus-worker`) is in the call, and `lib/lib.ts` gains `forceSubagentCallAsync()`.

Three constraints shaped it, all verified against pi-subagents source:

- **`async` is a run-level parameter.** It exists only as a top-level field (`ParallelTaskSchema` and `ChainItem` have none; `ChainItem` even sets `additionalProperties: false`), and the mode is derived there: `requestedAsync = effectiveParams.async ?? asyncByDefault`, `runsForeground = … (dispatchParams.async ?? asyncByDefault) !== true`. An item-level `async` on `tasks[]`/`chain[]` is read by nothing — writing one is a silent no-op, so the pin writes the **root** object. One call has one mode, so a composite carrying a Chorus agent is pinned as a whole.
- **`defaultAsync` is not a guarantee.** Frontmatter `async: "true"` only fills in when the caller omitted `async` *and* only in single mode — a caller's `async: false` wins. Hence a pin rather than a default.
- **`clarify` must be deleted, not falsified.** Two independent reasons, both in 0.66.0 and 0.70.0: it defeats async in the executor (`effectiveAsync = requestedAsync && clarify !== true`), and the public normalizer rejects a dispatch whenever the property is merely **defined** — `params.clarify !== undefined` in `src/extension/public-execution.js:91`, *"Public workflowScript execution does not support clarify UI."* — so assigning `false` would have turned a `clarify: true` call into a hard pre-dispatch rejection instead of a pin. The hook now issues `delete obj.clarify` (the third bullet in this list was wrong until the review round below). What remains true from the original reasoning: this mirrors the package's own `applyForceTopLevelAsyncOverride()`, which additionally honors `foregroundOnly` — a Chorus agent cannot run foreground at all, so this deliberately does not.

The override is **not silent**: when the caller explicitly passed `async: false`, the extension notifies once. The bundled subagent ignores both flags (its child is a separate `pi` process that does load extensions), so this is a no-op there — `extensions/subagent/**` is untouched in this PR.

## 2. Wording — de-mislead, and fix two false statements

- `skills/{review,develop,yolo}`, the injected Quick Reference, the three nudges and the four agent descriptions no longer mandate blocking; they say to spawn the reviewer and wait for its VERDICT, and note that the verdict is the *comment*, not the call's return value.
- `yolo` and `develop` claimed "The `subagent` tool is **blocking** … There is no async spawn, no `agentId`" — false for nicobailon, whose async support this package's own code depends on. Both now describe the two implementations.
- Residual "closes the session when the tool call returns" statements (worker body, Quick Reference, README ×2) now name both close paths.

## 3. Review round 2 (03c00e5e) — corrections

The review found both claims added in the first commit wrong at the public-dispatch boundary; both are fixed and mutation-checked.

- **`clarify` is removed, not set to `false`** (`lib/lib.ts`), for the reason in §1. Coverage exercises incoming `true` **and** `false`, on the single-child and composite shapes, plus a new assertion that loads the real `normalizePublicSubagentExecution()` from an installed pi-subagents (env override → `~/.pi/agent/npm` → project `node_modules`; skips where the package is absent) and checks the rewritten call is accepted with `params.async === true` and no surviving `clarify`. README §"Reviewers and workers are pinned" was synchronized.
- **The composite shape is bundled-only, and the new async-receipt promise for it was wrong.** Under nicobailon the public tool rejects top-level `tasks`/`chain` (`Legacy top-level chain and parallel inputs were removed; use workflowScript.`) *before* dispatch, so `subagent({ tasks: [...] })` never reaches the async path the old wording promised. `develop` / `yolo` / the injected Quick Reference / `agents/chorus-worker.md` / the README now split the two implementations explicitly: bundled = one composite call (blocking), nicobailon = one single dispatch per worker, issued together for the same wave (async, run receipt per call). The `workflowScript` alternative is documented with its catch — children created inside a script are invisible to the session hook, so they get no Chorus session.
- **Stale dispatch/close tool names removed** (not requested, same class of defect): `skills/{chorus,orchestrate,quick-dev,review}` and `docs/CONNECT_PI.md` still instructed `subagent_spawn` / `subagent_manage close`, leftovers from the narumitw → official `subagent` migration (see the archived `optimize-pi-plugin-npm-parity` design). They now use the `subagent` tool and the no-handle lifecycle, and static check **A6** guards both names so they cannot return. Say the word if you would rather see this split into its own PR.

## Verification

`test/all.sh` — static **41/0**, lib **77/0**, extension events **29/0**; `node scripts/validate-package.mjs` passes.

Mutation-checked on `/tmp` copies (repo untouched, cleaned up) — reverting any part fails the suite:

| mutant | result |
|---|---|
| baseline | static 41 / unit 106 pass, 0 fail |
| root write inverted | static clean, 11 fail |
| pin call removed | static clean, 7 fail |
| **pin regressed to an item-level write** (the bug found in review round 1) | static clean, 4 fail |
| worker half of the predicate dropped | static clean, 1 fail |
| override notification removed | static clean, 1 fail |
| **`clarify` assigned `false` instead of deleted** (review round 2) | static clean, 4 fail (2 lib incl. the public-boundary assertion, 2 extension-event) |
| **a skill reintroduces `subagent_spawn`** | static A6 fails |

The premise behind pinning the worker was measured rather than assumed: a probe agent that declares **no** `tools` allowlist (so no diagnostic can fire) was dispatched foreground and reported its entire tool set as `read, bash, edit, write, bg_wait, contact_supervisor` — no `mcp`, no `chorus_*` — and could not post a comment; a Chorus read-back confirmed no comment landed.

Re-checked on **pi-subagents 0.70.0** (the current release, which post-dates this branch): the assumptions behind the pin still hold — `ParallelTaskSchema` / `ChainItem` still expose no `async`, the mode is still derived from the top-level field, `clarify` still defeats async (and, as review round 2 established, any *defined* `clarify` is rejected outright), and there is still no enforced per-agent mode. The pin was also exercised end to end in an interactive session (override applied, notification rendered, reviewer posted its VERDICT). Details in the review-round comment below.

## Known gaps (documented in README, not fixed here)

The `tool_call` hook only reaches calls whose items expose `{agent, task}` and whose launch goes through the `subagent` tool. Not covered: reviewers nested in a chain step's `parallel[]` / dynamic `expand` fanout; a chain step that names only `agent` (its `task` defaults to `{previous}`); children created inside a `workflowScript`; `{action:"resume"}` replays; non-`subagent` spawn paths. Every shape the skills and nudges instruct is covered. Closing the rest means widening the enumeration (a new behaviour surface for the worker injection too), so the durable fix is upstream.

## Upstream

Filed as **nicobailon/pi-subagents#2347**: extend #2274's parent-registry sharing from provider **models** to the **tool** registry — then agents that need MCP no longer need `async: true`, and this pin (and the wording) become unnecessary. That issue is the root-cause path; this PR is the interim measure so that a Chorus agent can never be dispatched into a mode where it cannot do its job.

One more reason this belongs in the instructions and not only in the README: on 0.70.0, layouts where Pi sits outside the extension's `node_modules` fail **every** foreground launch before the child starts (upstream #2346: `Cannot find package '@earendil-works/pi-coding-agent' imported from …/child-session.js`). Because the pin rewrites the mode before launch, Chorus agents keep working there; without it, "use the blocking `subagent` tool" would now be a hard startup failure rather than a tool-availability failure.


expoli:
## Verified on pi-subagents **0.70.0**

This branch was written against 0.66.0, so here is a re-check on the current release — including an end-to-end run — since the mode decision is exactly the area 0.69/0.70 touched (#2289, #2301).

### The assumptions still hold on 0.70.0

Checked against the published package (`src/**/*.js` in 0.70.0):

| assumption | 0.70.0 |
|---|---|
| `async` is run-level only | `ParallelTaskSchema` / `ChainItem` still declare no `async` |
| the mode is derived from the top level | `runs/foreground/subagent-executor.js:6963` |
| `clarify: true` still defeats async | `:6528` — `effectiveAsync = requestedAsync && params.clarify !== true` |
| there is still no enforced per-agent mode | `agents/agents.js:1862-1865` — `frontmatter.async` feeds `defaultAsync` only |
| `fallbackModels` is now a hard error | `agents/agents.js:707` (overrides) and `:1838` (frontmatter) — so downstream settings that still carry it must drop it before upgrading |

### End-to-end, interactive session

pi 0.86.0 · pi-subagents 0.70.0 · this branch installed from a local path.

- dispatched `chorus-task-reviewer` with an explicit **`async: false`**;
- the extension rewrote the mode before launch: the call returned `Async: chorus-task-reviewer [2f51896e…]`, the run settled `complete`, `events.jsonl` carried no error, `runner.stderr.log` was empty;
- the notification rendered as a UI toast (only observable interactively): *"Chorus agents need ambient `mcp`/`chorus_*` tools, which a foreground child does not have — this dispatch was pinned to the background (async) path."*
- the reviewer ran with its MCP tools and posted its `VERDICT` comment — verified by reading the task's comments back, not from the child's own report.

### Side effect worth knowing: this also shields a 0.70.0 foreground regression

On 0.70.0, in a layout where Pi lives outside the extension's `node_modules` (the common `~/.pi/agent/npm` + global/npm install), **every** foreground launch now fails before the child starts:

```text
Cannot find package '@earendil-works/pi-coding-agent' imported from
…/pi-subagents/src/runs/shared/child-session.js
```

Reproduced here 2/2 with a plain built-in `scout` + `async: false`, while the same call with `async` succeeded and returned the expected content — upstream issue #2346. Because the pin rewrites the mode before launch, Chorus agents keep working on 0.70.0; without it, "use the blocking `subagent` tool" would now be a hard startup failure rather than a tool-availability failure. That is a second reason this belongs in the instructions and not only in the README — and, FWIW, a second reason for #2347: a foreground child that could share the parent's tool registry would need no pin at all.

packages/chorus-pi/skills/develop/SKILL.md
```diff
@@ -347,12 +349,12 @@ chorus_submit_for_verify({ taskUuid: "<my-task-uuid>", summary: "..." })
 
 **Wave-based execution (recommended):**
 1. `chorus_get_unblocked_tasks` — find ready tasks
-2. Dispatch a `chorus-worker` per ready task in ONE blocking `subagent({ tasks: [...] })` call (max 8; batch if more). The call returns when the whole wave has finished (each worker at `to_verify`).
+2. Dispatch a `chorus-worker` per ready task in ONE `subagent({ tasks: [...] })` call (max 8; batch if more). Wait for the whole wave to settle (each worker at `to_verify`) — under nicobailon `pi-subagents` the call returns a run receipt, so wait on the run notification / `bg_wait`.
```
ChenNima: This newly added nicobailon receipt claim is incorrect for pi-subagents 0.66.0 and 0.70.0: the public normalizer rejects top-level tasks/chain before dispatch (`Legacy top-level chain and parallel inputs were removed; use workflowScript.`). Please separate bundled parallel examples from nicobailon guidance, e.g. multiple independent single dispatches, and make the equivalent yolo/README wording consistent.

packages/chorus-pi/lib/lib.ts
```diff
@@ -161,6 +169,51 @@ export function subagentTaskItems(input: unknown): SubagentTaskItem[] {
...
+ * @returns true when the caller had explicitly asked for foreground
+ * (`async: false`) and was overridden — so the caller can surface the override
+ * instead of silently changing the dispatch mode.
+ */
+export function forceSubagentCallAsync(input: unknown): boolean {
+  if (!input || typeof input !== "object") return false;
+  const obj = input as Record<string, unknown>;
+  const overrodeExplicitForeground = obj.async === false;
+  obj.async = true;
+  if (obj.clarify === true) obj.clarify = false;
```
ChenNima: Setting clarify to false does not clear it for the public API: pi-subagents 0.66.0 and 0.70.0 reject whenever `params.clarify !== undefined`, including false. This is confirmed by the actual public normalizers and a real 0.70.0 host call. Please delete the obsolete property for pinned calls and test both incoming true and false, rather than asserting false remains on the input.

ChenNima (changes requested):
Requesting changes on `6e1e13ce83f837f6c459efa6b5b25aab5e53095a`.

The root-level async pin is sound, but two compatibility issues need addressing before the shipped instructions and override behavior match the supported public tool interface:

1. **Correct the nicobailon parallel-dispatch instructions.** `develop/SKILL.md:352` and `yolo/SKILL.md:339,367` now state that `subagent({ tasks: [...] })` returns an async receipt under pi-subagents. In both published versions **0.66.0 and 0.70.0**, `normalizePublicSubagentExecution()` instead rejects top-level `tasks`/`chain` with `Legacy top-level chain and parallel inputs were removed; use workflowScript.` The legacy examples predate this PR, but the explicit promise of nicobailon async receipts is newly added here. Separate the bundled examples from nicobailon guidance; multiple independent single dispatches preserve the current hook's session injection. A workflowScript alternative needs to account for the hook's documented inability to enumerate its children.

2. **Remove the obsolete `clarify` property instead of assigning false.** `lib/lib.ts:213` changes `clarify:true` to `false`, but both versions reject any defined `clarify` before dispatch. Delete the property for the calls being pinned, cover both incoming true and false in the tests, and synchronize the README's explanation at lines 173–176.

Validation: static checks 40/0, direct library/event tests 104/0, and package validation passed. Both dependency versions' actual public normalizers reproduce the above rejections; a real Pi 0.85.1 / pi-subagents 0.70.0 call also confirmed the `clarify:false` rejection. Reviewer and bundled probes succeeded. The 0.70.0 worker demonstrated MCP access and correct session closure, but its no-edit diagnostic task ultimately failed the upstream implementation completion guard; that is a probe limitation, not a reason to reject this PR.

Please update the instructions and override, add regression coverage at the public-dispatch boundary, and rerun the relevant checks. The disclosed out-of-scope nested/script/resume paths do not need to be expanded to resolve these two requests.

packages/chorus-pi/skills/develop/SKILL.md
expoli: Fixed in 03c00e5e — the async-receipt promise for a composite call was wrong, and it is gone from all four surfaces you named (plus two more that carried the same claim).

**Separating bundled examples from nicobailon guidance**

- `skills/develop/SKILL.md` §Parallel Sub-Agent Integration now has an implementation table: bundled = one composite call (`tasks: [...]` / `chain: [...]`), blocking; nicobailon = **one single dispatch per worker**, issued together for the same wave, async by default. The "three modes" note says the composite modes are the **bundled** subagent's schema, quotes the rejection (`Legacy top-level chain and parallel inputs were removed; use workflowScript.`) and records the hook's inability to enumerate `workflowScript` children, so a nicobailon reader is steered to one call per child instead. The Team Lead example, the wave step in "Handling Task Dependencies", and the chain-mode line were rewritten the same way (chain is now marked bundled-only, with the `workflowScript` caveat).
- `skills/yolo/SKILL.md` §Phase 3 leads with the same split, the pseudocode shows both shapes with the composite labelled bundled-only, and a note after the loop says not to dispatch Chorus workers from a `workflowScript` (no session injection for those children).
- `README.md`: the "Subagent run modes" section states that the bundled tool is the only one that takes a composite call and that nicobailon accepts one child per call (quoting the rejection, noting 0.66.0/0.70.0); the coexistence table row and the `workflowScript` bullet carry the same qualification.
- Also updated for consistency: the injected Quick Reference in `extensions/chorus.ts` and `agents/chorus-worker.md`'s description, which both implied composite dispatch in general terms.

**Regression coverage**: the `subagent_spawn` / `subagent_manage close` names you did not ask about turned out to be stale too (three skills still instructed them; this port's only dispatch tool is `subagent`), so I replaced them with the `subagent` tool + no-handle lifecycle in `skills/{chorus,orchestrate,quick-dev,review}` and `docs/CONNECT_PI.md`, and added both names to static check A6 so they cannot come back (mutant-checked: reintroducing one fails A6). That widens the diff by those five files; say the word if you would rather have it split into its own PR.

**Re-ran**: static 41/0, lib 77/0, ext-events 29/0, `scripts/validate-package.mjs` passes. The wording itself stays review-verified rather than test-verified — I did not add a string-matching guard for prose.

packages/chorus-pi/lib/lib.ts
expoli: Fixed in 03c00e5e — thanks, both were right and I had the wrong compatibility claim.

**What changed**

- `lib/lib.ts`: the pin now **deletes** `clarify` (`if ("clarify" in obj) delete obj.clarify`) instead of assigning `false`. Reproduced your finding against the installed 0.70.0 normalizer: `{agent, task, async: true, clarify: false}` → `Public workflowScript execution does not support clarify UI.` (same check rejects `true`; 0.66.0 uses the identical `params.clarify !== undefined` guard at the same place in `normalizePublicSubagentExecution`). The doc comment now states that any *defined* value is fatal, so `delete` is the only safe rewrite; the reason `clarify: true` had to go (it defeats async in the executor) is kept as the secondary note.
- Tests cover **both** incoming values, on both the single-child and composite shapes: `test/lib.test.ts` asserts `"clarify" in input === false` for `true` and `false`; `test/ext-events.test.ts` drives the real hook with `clarify: false` and asserts the property is gone. A new assertion loads the **real** `normalizePublicSubagentExecution` from an installed pi-subagents (env override → `~/.pi/agent/npm` → project `node_modules`, skipped when absent), asserts the unpinned `clarify` call is rejected, then asserts the pinned call is `ok: true` with `params.async === true` and no surviving `clarify`.
- README §"Reviewers and workers are pinned" now says *removing* the `clarify` property (`delete`, not `clarify: false`) and names the normalizer + the reason `false` is equally fatal.

**Re-ran**: static 41/0, lib 77/0, ext-events 29/0, `scripts/validate-package.mjs` passes. Mutation-checked on a `/tmp` copy: reverting the `delete` back to `clarify = false` fails the two lib tests (deletion + boundary) and both ext-events clarify cases, so the coverage is real rather than incidental.

expoli:
@ChenNima both review items are addressed in **03c00e5e** (details in the two thread replies above) — requesting a re-review, as I don't have push access on the base repo to set the reviewer via the API.

**1. `clarify` is deleted, not falsified** (`lib/lib.ts`) — your finding reproduced against the installed 0.70.0 normalizer: `{agent, task, async: true, clarify: false}` → `Public workflowScript execution does not support clarify UI.` Coverage now exercises incoming `true` **and** `false` on both the single-child and composite shapes, plus an assertion that loads the real `normalizePublicSubagentExecution()` from an installed pi-subagents (env override → `~/.pi/agent/npm` → project `node_modules`; skips where absent) and checks the pinned call is accepted with `params.async === true` and no surviving `clarify`.

**2. The composite async-receipt promise is gone**, and the two implementations are separated everywhere: `tasks: [...]` / `chain: [...]` is labelled the **bundled** subagent's schema; under nicobailon a wave is **one single dispatch per worker**, issued together (its public tool rejects top-level `tasks`/`chain` before dispatch), with the `workflowScript` caveat (children invisible to the session hook → no Chorus session). Updated in `develop` (new implementation table + examples + wave step + chain line), `yolo` (paragraph + pseudocode + a note after the loop), the README (run modes / coexistence table / `workflowScript` bullet / pin bullet), the injected Quick Reference and `agents/chorus-worker.md`.

**Also in this push** (same class of defect, tell me if you want it split out): `skills/{chorus,orchestrate,quick-dev,review}` + `docs/CONNECT_PI.md` still instructed `subagent_spawn` / `subagent_manage close`, leftovers from the narumitw → official `subagent` migration. They now use the `subagent` tool and the no-handle lifecycle, and static check **A6** guards both names.

**Re-ran**: static 41/0, lib 77/0, ext-events 29/0, `scripts/validate-package.mjs` passes; CI `test` green on 03c00e5e. Mutation-checked on `/tmp` copies: `clarify = false` again → 4 failures (incl. the boundary assertion), a reintroduced `subagent_spawn` → A6 fails.

packages/chorus-pi/agents/chorus-worker.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: chorus-worker
-description: General-purpose Chorus implementer subagent that claims and completes ONE Chorus task end-to-end via the develop workflow. Dispatch it via the blocking subagent tool (single or parallel mode) for wave-based execution.
+description: General-purpose Chorus implementer subagent that claims and completes ONE Chorus task end-to-end via the develop workflow. Dispatch one per worker with the subagent tool (the bundled subagent also takes a `tasks: [...]` composite) for wave-based execution, and wait for the run to settle.
```
ChenNima: [P1] Quote this YAML description. The newly added `tasks: [...]` has a colon-space inside a plain scalar; Markdown backticks do not escape it. Actual Pi 0.85.1 bundled dispatch now throws `Nested mappings are not allowed in compact mappings` during agent discovery, before launching the worker. Pi's own parseFrontmatter parses 6e1e13ce successfully but rejects 03c00e5e; quoting the description in memory restores parsing. Please quote it or use a block scalar and add a real frontmatter-parsing regression check. Because discovery parses the full bundled agents directory, other bundled agents can also be blocked.

ChenNima (changes requested):
The two previously requested changes are fixed in 03c00e5e: obsolete `clarify` is removed, and the docs distinguish bundled composites from nicobailon's single-dispatch API. However, actual Pi runtime testing found a new blocking regression in the updated worker description.

With Pi 0.85.1 and the bundled subagent extension loading the shipped agents, the first `chorus-worker` dispatch fails before launching a worker:

```
Nested mappings are not allowed in compact mappings at line 2, column 14:
description: General-purpose Chorus implementer subagent ...
```

The newly added backticked `tasks: [...]` contains a colon followed by a space inside an unquoted YAML scalar. Markdown backticks do not quote YAML. I independently reproduced this with Pi's own `parseFrontmatter`: the previous head 6e1e13ce parses, while current head 03c00e5e throws; quoting the description in memory makes it parse.

Please quote the whole description or use a YAML block scalar, and add a regression check that parses the shipped agent frontmatter using the runtime parser. `discoverAgents` parses the whole bundled directory before choosing an agent, so this can also prevent other bundled agents from being dispatched.

The nicobailon 0.70.0 path did successfully run two real workers concurrently: both made fixture edits, passed their tests, used ambient Chorus MCP, and completed with exit code 0. The legacy clarify true/false inputs were both removed before dispatch, and both worker sessions closed. That success does not cover the bundled parser failure. The PR already targets develop; merging remains blocked by this runtime regression.

expoli:
Both requested changes are in f6c63a21.

**Quoted description.** `agents/chorus-worker.md` now wraps the description in double quotes — the backticked `tasks: [...]` is literal inside a double-quoted scalar, and the text contains no quote/backslash characters needing escapes. (Considered a `>-` block scalar; quoting was the smaller diff.)

**Runtime-parser regression check.** New `test/agents.test.ts`, wired into `test/all.sh` as its own Layer B step: parses every shipped `agents/*.md` with the same runtime `parseFrontmatter` and applies `loadAgentsFromDir`'s acceptance line — `name`/`description` must be non-empty strings, `name` must equal the file stem, body must be non-empty. A frontmatter that cannot load now fails offline instead of at first dispatch.

Mutation-checked end to end: reverting the quotes fails the suite (4 pass / 1 fail, the failure carrying the parser's "Nested mappings" message). With the fix restored, `test/all.sh` is green (static 41/0, lib 77/0, extension events 29/0, agents 5/0) and `node scripts/validate-package.mjs` passes (4 agents).

One incidental fix in the same file, because the regression check would have been toothless without it: `all.sh` captured the bun layers' status *after* a pipe to `tail`, so `$?` held tail's always-0 status — a failing bun layer never failed the suite (`b=$?` was checking the wrong command). The status is now captured via command substitution before the display pipe. Verified: with the quotes reverted, `all.sh` exits 1 with `B-agents exit=1` instead of silently printing PASSED.

ChenNima (approved):
Verified f6c63a21. The YAML description is now quoted, and the new runtime-parser tests cover all shipped agent files. Restoring the prior broken description in an isolated fixture makes the regression test fail. Local validation: 41 static checks and 111 tests passed (77 helper tests including the real pi-subagents 0.70.0 public normalizer, 29 extension-event tests, 5 agent-frontmatter tests); package validation passed.

Actual Pi 0.85.1 integration: both bundled subagent and external pi-subagents 0.70.0 completed two real implementation workers plus an independent reviewer, all six child runs successful with exit code 0. Both parents exited 0; ambient MCP writes worked and all four worker sessions closed. The bundled launch regression is resolved. Existing CI is green; the Pi-specific suite above was explicitly run locally. No remaining blocking findings. Approved for develop under the requester's merge authorization.

